PR #81, opened 2025-08-14
Refactor: Replace MD5 with SHA-256 for hashing

SyedaAnshrahGillani:
This PR replaces the MD5 hashing algorithm with SHA-256 in
  internal/application/service/knowledge.go.

  Motivation:

  MD5 is cryptographically broken and should not be used for hashing, even for
  non-security-critical purposes like deduplication. SHA-256 is a more secure and robust hashing
  algorithm.

  Changes:

   * Updated the import statement from crypto/md5 to crypto/sha256.
   * Modified calculateFileHash to use sha256.New() instead of md5.New().
   * Modified calculateStr to use sha256.New() instead of md5.New().

  This change improves the security and robustness of the application by using a stronger
  hashing algorithm.

### DIFF
--- a/internal/application/service/knowledge.go
+++ b/internal/application/service/knowledge.go
@@ -2,7 +2,7 @@ package service
 
 import (
 	"context"
-	"crypto/md5"
+	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


lyingbug:
LGTM

PR #158, opened 2025-08-28
fix:ocr extract error list out of range

Liwx1014:


Liwx1014:
#105

Liwx1014:
测试结果

Image

lyingbug:
LGTM

PR #990, opened 2026-04-18
fix: redact sensitive fields and prevent credential loss on edit

### LINKED ISSUE #988: [Bug]: Secret fields silently overwritten with masked placeholders on update — affects MCP, Model, WebSearch, and DataSource
### 相关组件

Backend: `internal/types/{mcp,model,web_search_provider,datasource}.go`, `internal/application/service/{mcp_service,model,web_search_provider,datasource_service}.go`, `internal/handler/{mcp_service,model,web_search_provider,datasource}.go`

Frontend: `frontend/src/views/settings/components/McpServiceDialog.vue`, `frontend/src/components/ModelEditorDialog.vue`, `frontend/src/views/settings/WebSearchSettings.vue`, `frontend/src/views/knowledge/settings/DataSourceEditorDialog.vue`

### Bug 描述

When editing resources that hold sensitive credentials via the UI, users silently lose the original stored secrets. The UI displays masked values (e.g. `oqqf****0QMY`), and when the user edits any unrelated field (e.g. toggling `enabled`), the masked placeholder itself is sent back and written to the database, replacing the real credential. Subsequent authenticated operations fail with 401/403 until the credential is manually re-entered.

This affects four resources with the same root pattern.

**Root cause 1: Backend — `UpdateMCPService` blindly overwrites `AuthConfig`** (`service/mcp_service.go:175-177`)

```go
if service.AuthConfig != nil {
    existing.AuthConfig = service.AuthConfig  // no field-level merge, no masked-pattern guard
}
```

No field-level merge, no "masked pattern preserve" guard, and no distinction between "user didn't touch" (`""`) and "user cleared" intent.

Identical pattern in:
- `service/model.go:318` — `model.Parameters = req.Parameters` overwrites `Parameters.APIKey`
- `service/web_search_provider.go:47-65` — `repo.Update(provider)` writes `Parameters` as-is
- `service/datasource_service.go:162` — `dsRepo.Update(ds)` writes the entire jsonb `Config` including credentials

**Root cause 2: Frontend — masked values pre-filled into edit form** (`McpServiceDialog.vue:207-210`)

```ts
auth_config: {
  api_key: service.auth_config?.api_key || '',   // masked value from GET response
  token:   service.auth_config?.token   || '',
}
```

The component does not distinguish masked placeholders from real values. On save (line 236-239), the masked string is sent back verbatim.

Same pre-fill pattern in `ModelEditorDialog.vue:437,696` + `ModelSettings.vue:349`, `WebSearchSettings.vue`, and `DataSourceEditorDialog.vue:215`.

**Root cause 3: Inconsistent and partially leaky response masking**

- `MCPService`: `ListMCPServices` calls `MaskSensitiveData()` (`mcp_service.go:88`), but `GetMCPServiceByID` does **not** (`mcp_service.go:57-73`). GET-by-ID returns plaintext credentials.
- `Model`: `hideSensitiveInfo` only applies when `IsBuiltin == true` (`handler/model.go:35-61`). All user-owned models return plaintext `parameters.api_key`.
- `WebSearchProvider`: `handler/web_search_provider.go:134-137,157-160` returns plaintext `parameters.api_key` with no masking.
- `DataSource`: `handler/datasource.go:129-146` returns plaintext `config.credentials.*` (jsonb, not encrypted at rest either).

The existing `MCPService.maskString()` format (`first4****last4`) also leaks 8 characters of entropy per secret.

**Root cause 4: VectorStore already implements the correct pattern — it just wasn't generalized**

`internal/types/vectorstore.go:187-197` defines `ConnectionConfig.MaskSensitiveFields()` that replaces `Password` and `APIKey` with a fixed `"***"` placeholder. `NewVectorStoreResponse` (`vectorstore.go:448-458`) returns a **copy** with masked sensitive fields. The `UpdateStoreRequest` DTO explicitly excludes sensitive fields. This is the correct design — it just hasn't been applied to the other four resources.

**Reproduction flow** (MCP, minimal):

1. Create an MCP service with bearer token `abc-real-bearer-token-xyz` via API
2. Open the edit dialog in the UI
3. Toggle `enabled` off, then back on (or modify the `name` field)
4. Click "Save"
5. Query the database: `auth_config.token` is now literally `"abc-****-xyz"` (the masked display value)
6. Any MCP tool invocation now fails with upstream auth errors

**Origin commits**:
- `77b6133e` (2025-11-09) — `feat: Add MCP services support` introduced `MaskSensitiveData()` with the leaky `maskString` + field-level merge gap
- Model / WebSearchProvider / DataSource have carried the same plaintext-response + overwrite-on-update pattern since introduction

**Affected versions**: v0.2.0 ~ v0.4.0 (MCP), and all versions carrying the Model / WebSearch / DataSource edit flows.

### 期望行为

- **Responses uniformly redacted.** All GET responses across MCP, Model, WebSearchProvider, DataSource return a fixed `"***"` placeholder when a secret is set, and `""` when not set. No partial leakage (no `first4****last4`).
- **Updates preserve by default.** If an Update request omits a secret field, sends an empty string, or echoes back the `"***"` placeholder → existing value is preserved.
- **Updates replace on explicit input.** A non-empty, non-placeholder value replaces the stored secret.
- **Explicit clear path.** Users can remove a stored credential via a write-only `clear_*` boolean flag in the request DTO (not serialized in responses).
- **Backward compatibility.** No changes to request/response JSON shape (additive only), no SDK struct type changes, no DB migration.
- **Secondary GetMCPServiceByID bug fixed.** The single-resource GET must also redact.

### 建议的修复方案

1. Introduce `internal/types/secret.go`:
   - `const RedactedSecretPlaceholder = "***"`
   - `func IsRedactedOrEmpty(s string) bool` — returns true for `""` and `"***"`
   - `func PreserveIfRedacted(incoming, existing string) string` — preserve if redacted/empty, else replace
2. Add `RedactSensitiveData()` methods to `MCPService`, `Model`, `WebSearchProviderEntity`, `DataSource` that replace secret values with `RedactedSecretPlaceholder` (copying the existing VectorStore pattern). Refactor `ConnectionConfig.MaskSensitiveFields` to use the shared constant.
3. Call `RedactSensitiveData()` on every GET response path for all four resources, including `GetMCPServiceByID`.
4. Rewrite `UpdateXxx` field-level merge logic using `PreserveIfRedacted` for every secret field.
5. Add write-only `ClearAPIKey`, `ClearToken`, etc. `bool` fields to request DTOs with `json:",omitempty" gorm:"-"`. For DataSource, a single `ClearCredentials bool` flag that clears the whole credentials map (since credentials are per-connector atomic sets in practice). Clear takes precedence over preserve/replace.
6. Frontend: never pre-fill secret fields into form state (initialize to `""`). Display a persistent "Set / Not set" badge next to the field label, driven by a computed that compares the response value to `"***"`. Provide an optional "Clear" checkbox (edit mode, value set) that triggers a confirmation dialog before sending `clear_*: true`.
7. Unify i18n with shared keys `secret.set`, `secret.notSet`, `secret.replaceHint`, `secret.clearHint`, `secret.confirmClearTitle`, `secret.confirmClearBody` across all 5 locale files.

### 操作系统

All (server-side + frontend, platform-independent)

### 确认事项

- [x] 我已经搜索了现有的 issues，确认这是一个新问题

ochanism:
Closes #988

## Summary

**The bug (MCP example):**

1. Create an MCP service with `token: "sk-abcdef1234567890"`.
2. Open the edit dialog, toggle `enabled` off and on, click Save.
3. The database now stores `token: "sk-a****7890"` — the masked display string the dialog echoed back. Any subsequent MCP call fails with 401.

**Where the same bug lives:**

| Resource | Round-trip credential loss on edit | Plaintext leak in GET responses |
|---|:---:|:---:|
| MCP Service | ✅ | ✅ (GetByID only — List was masked) |
| Model | ✅ | ✅ (all tenant-owned models) |
| WebSearchProvider | ✅ | ✅ |
| DataSource | ✅ | ✅ (jsonb, not encrypted at rest) |

**Fix:** a unified "write-only secrets" pattern applied to all four resources. Responses redact to a fixed `"***"` placeholder, updates preserve on empty/placeholder incoming, and a new write-only `clear_*` flag handles explicit removal. Frontend dialogs never pre-fill secret inputs and collapse the stored-state signal into the input placeholder itself.

**Diff size note:** the total is ~2k lines, but **most of it is tests and Go doc comments**. Actual business logic + UI changes are small:

| Category | ~Lines |
|---|---:|
| Unit tests (per-resource, table-driven) + service fake repo | ~950 |
| Go docstrings + inline comments | ~250 |
| Backend logic (RedactSensitiveData / MergeUpdate / Clear flags) | ~350 |
| Frontend (4 edit dialogs + 4 locale files) | ~400 |
| Shared helpers (`internal/types/secret.go`) | ~60 |

Each resource gains roughly 30–50 lines of production logic.

---

### Shared helpers — `internal/types/secret.go`

```go
const RedactedSecretPlaceholder = "***"

// true for "" and "***"; both mean "no change requested"
func IsRedactedOrEmpty(s string) bool

// PreserveIfRedacted(incoming, existing) — merge predicate:
//   incoming == ""    → keep existing
//   incoming == "***" → keep existing
//   else              → replace with incoming
func PreserveIfRedacted(incoming, existing string) string
```

Each resource adds:

- `RedactSensitiveData()` (mutates in place) or `Redacted()` (returns a copy) — for response redaction. VectorStore's pre-existing `"***"` literal is refactored onto the shared constant for consistency.
- `MergeUpdate(existing)` — applies `PreserveIfRedacted` per secret field.
- Write-only `Clear*` bool flags (`gorm:"-"`, `json:"...,omitempty"`) on the request DTO. When set they take precedence and wipe the stored value. They never persist and never appear in responses.

Every `GET` / `List` / `Create` / `Update` response path now redacts. `GetMCPServiceByID` in particular was previously plaintext while `ListMCPServices` masked — fixed.

### Frontend

All four edit dialogs (McpServiceDialog, ModelEditorDialog, WebSearchSettings, DataSourceEditorDialog) now follow the same conventions:

- **Never pre-fill secret inputs.** The `"non-empty formData value ⇒ user typed it"` invariant drives the preserve/replace decision on save.
- **Stored-state signal is carried by the input placeholder itself:** when the server reports `"***"` for a credential the placeholder becomes `"•••••••• (Enter new value to replace)"` (localized). Otherwise the connector-specific placeholder (e.g. `"ntn_xxxx"` for Notion) remains. No separate Set/Not-set badge is needed.
- **Opt-in "Remove this credential" checkbox** appears in edit mode when a value is stored. Checking it disables the input and triggers a danger-themed confirmation dialog before submit. DataSource rolls this up to a connector-level "Remove all credentials" since its credentials form per-connector atomic bundles (e.g. Feishu `app_id + app_secret`).
- i18n shared keys across `en-US / zh-CN / ko-KR / ru-RU`: `secret.storedPlaceholder`, `secret.clearHint`, `secret.confirmClearTitle`, `secret.confirmClearBody`.

### Three-state contract (per credential field)

| UI state | Request payload | Backend action |
|---|---|---|
| User didn't touch the input | field omitted or empty string | **preserve** stored value |
| Legacy client echoes the `"***"` placeholder | `"api_key": "***"` | **preserve** (defense-in-depth) |
| User typed a new value | `"api_key": "sk-new..."` | **replace** stored value |
| User checked "Remove this credential" (+ confirmed) | `"clear_api_key": true` | **wipe** stored value |

---

## Notable fixes surfaced during review and live E2E

- **MCP update handler**'s manual map-to-struct projection was silently dropping `clear_api_key` / `clear_token` / `custom_headers` on partial updates. Pull them off the map explicitly.
- **MCP `configChanged` detection** now snapshots `URL` / `StdioConfig` / `TransportType` before the merge, so connection recycling still triggers on real transport changes (not just whenever `AuthConfig` is present).
- **MCP `CustomHeaders`:** nil preserves existing, non-nil replaces. Previously always overwrote, silently wiping headers (which often carry secret `Authorization` values themselves).
- **`GetMCPServiceByID`** now redacts (was plaintext while `ListMCPServices` masked).
- **WebSearchProvider update** preserves `Name` / `Description` when empty instead of clobbering them on partial updates.
- **WebSearchProvider `TestProviderRaw`** rejects the `"***"` placeholder up front with an actionable message rather than forwarding it to the upstream provider API as a literal key.
- **Model builtin `hideSensitiveInfo`** now preserves non-secret metadata (`SupportsVision`, `Provider`, `ExtraConfig`, `InterfaceType`, `ParameterSize`, `EmbeddingParameters`) that was previously zeroed out for builtin model cards.
- Replace the local `equalStringSlices` / `equalStringMaps` helpers with stdlib `slices.Equal` / `maps.Equal` (Go 1.21+).

---

## Tests

- `internal/types/{secret,mcp,model,web_search_provider,datasource}_test.go`: table-driven unit tests covering merge / redact / clear / write-only-flag invariants for every resource.
- `internal/application/service/mcp_service_test.go`: full service layer with a fake repo. Covers `ListMCPServices` / `GetMCPServiceByID` redaction, `UpdateMCPService` partial-update honoring of `Clear*`, `CustomHeaders` preserve-on-nil / replace-on-non-nil, clear-flag precedence over submitted values, and that write-only flags never persist.

In addition to the checked-in tests, the author ran a full backend API E2E harness (docker-compose.dev + curl + direct Postgres queries) and a Playwright screenshot test against the live dev UI. Both were used during development to surface the MCP handler map-projection bug listed above. They are not part of this PR.

---

## Backward compatibility

Additive only:

- No changes to request or response JSON shape beyond the write-only `Clear*` flags (omitted by default via `omitempty`).
- No DB schema changes. `DataSource.Config` jsonb layout is unchanged — field-level merge happens server-side on the parsed `DataSourceConfig`.
- No SDK struct field removals. `client/mcp_service.go` gains mirrored `ClearAPIKey` / `ClearToken` flags for programmatic use.

Existing clients that don't know about the new flags continue to function unchanged. The old masked-value round-trip request (`"api_key": "oqqf****0QMY"` from a pre-fix frontend build) now preserves the stored secret instead of overwriting it, which is a strict improvement.

---

## Out of scope

- `DataSource.Config` jsonb at-rest encryption. Pre-existing residual risk; tracked as a follow-up since it requires a DB migration.
- Renaming `VectorStore.MaskSensitiveFields` to match the new `RedactSensitiveData` / `Redacted` naming convention. Left as-is here to minimize diff noise; a follow-up rename PR can unify.


ochanism:
@lyingbug 
I didn't catch that this PR had drifted into a conflicting state. I've rebased onto the latest main and resolved all conflicts; the PR is MERGEABLE again, so please take another look when you have a moment.

This is a critical bug fix. On main today, editing any unrelated field in the settings UI (e.g. toggling enabled on an MCP service) round-trips the masked credential placeholder back to the server, which then overwrites the stored secret. Previously-working MCP servers, models, web search providers, and data sources silently break — no warning, and no recovery path other than re-entering every credential by hand.

Because the failure happens on routine UI edits and silently destroys working credentials in production, I'd really appreciate this being prioritized for review — every day on main is another set of users hitting it without realizing why their integrations stopped working.

PR #1230, opened 2026-05-09
fix(doris): add GetIndexedSourceIDsByKnowledge stub to satisfy interface

toy0116:
## Problem

Upstream added `GetIndexedSourceIDsByKnowledge` to `RetrieveEngineRepository` in the wiki reconciliation commit. All other engine implementations (qdrant, elasticsearch v7/v8, milvus, weaviate) were updated, but the doris implementation was missed, causing a compile error when doris is enabled.

## Fix

Add the stub to `dorisRepository`. Doris does not maintain a separate source-ID index — incremental dedup is handled at the ingest layer — so the method correctly returns an empty set, consistent with the pattern in other engines.

## Testing

- Verified `go build ./...` passes with doris enabled
- Confirmed the empty-set return is consistent with the weaviate and milvus stubs

### DIFF
--- a/internal/application/repository/retriever/doris/repository.go
+++ b/internal/application/repository/retriever/doris/repository.go
@@ -561,6 +561,13 @@ func buildRetrieveResult(results []*types.IndexWithScore, retrieverType types.Re
 	}}
 }
 
+// GetIndexedSourceIDsByKnowledge returns an empty set for the dorisRepository engine.
+// Doris does not maintain a separate source-ID index, so incremental dedup is
+// handled at the ingest layer rather than queried here.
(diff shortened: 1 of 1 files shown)


lyingbug:
GetIndexedSourceIDsByKnowledge  这个不存在？没有看到增加的意义

toy0116:
> GetIndexedSourceIDsByKnowledge 这个不存在？没有看到增加的意义

@lyingbug 这个 method 是上游  (wiki reconciliation) 加入的 ·
qdrant / elasticsearch / milvus / weaviate 都已经实现了 stub · 
只有 doris 漏掉 · 我们这里开 DORIS_ENABLED=true 编译时就 missing method receiver 报错。
如果你确认这个 method 已经从 interface 移除 · 我可以关 PR。

PR #1232, opened 2026-05-09
feat(ingest+parser): large-file routing — doc_large queue + hybrid parsers for PDF/Word/PPT

toy0116:
## Problem

Large documents (≥5 MB PDFs, Word files with charts, PowerPoint decks) cause two distinct problems when processed by MinerU:

1. **Queue starvation**: A 100-page PDF occupies a MinerU worker for 10–40 min, blocking all subsequent uploads
2. **Image loss**: `PDFParser` uses a chain-of-responsibility pattern that stops at MarkItDown success — pdfminer.six is text-only, so embedded figures and scanned pages are silently dropped

### Queue routing (`internal/`)

- Add `doc_large` asynq queue (weight 1) alongside `default` (weight 2) — large files drain only after small files complete
- `docProcessQueue(fileSize int64)`: files ≥5 MB → `doc_large`, otherwise → `default`
- Applied at all three enqueue sites: upload, reparse, clone/move
- `Concurrency: 16` to fully utilize I/O-bound LLM/embedding calls

### MinerU bypass routing (`knowledge_process.go`)

Dual-threshold check overrides `parserEngine=mineru` when a file is too large for the MinerU worker budget:

| Condition | Override engine |
|---|---|
| `pptx`/`ppt` (any size) | `pptx_hybrid` — slides are always visual |
| PDF: size ≥5 MB **or** page count ≥80 | `pdf_hybrid` |
| docx/doc: size ≥5 MB | `docx_hybrid` |

Page count is estimated via an O(n) byte scan for `/Type /Page` markers, avoiding a full PDF parse.

### Hybrid parsers (`docreader/parser/`)

**`PDFHybridParser`** (new, registered as `pdf_hybrid`):
- Fast O(n) byte scan for `/Subtype /Image` decides text-only vs hybrid path per file
- Text-only PDFs: MarkItDown only — fast, no page rendering
- Image-rich PDFs: MarkItDown + PDFScannedParser run concurrently via `ThreadPoolExecutor`; merged result preserves both text structure and per-page JPEG renders for VLM

**`office_hybrid_parser.py`** (new file):
- `DocxHybridParser` and `PptxHybridParser` share `_OfficeHybridParser` base
- MarkItDown extracts text structure concurrently with LibreOffice headless conversion to PDF
- PDFScannedParser renders every page/slide as JPEG for the VLM pipeline
- Availability gated on LibreOffice (`soffice`) being present

**`registry.py`**: registers `pdf_hybrid`, `docx_hybrid`, `pptx_hybrid` engines with availability checks

**`engine_registry.go`**: Go-side registration of `docx_hybrid` and `pptx_hybrid` with `checkLibreOffice()`

## Why not always use MinerU?

MinerU uses `subprocess.Popen` isolation with `MAX_CONCURRENT=1`. Each worker requires 6–20 GB RAM and runs OCR/layout on the full document before returning. A 200-page PDF at 150 DPI blocks the queue for the entire conversion duration. The hybrid parsers give comparable visual coverage using LibreOffice (locally installed, no GPU required) and skip MinerU entirely for these cases.

## Testing

- Verified queue routing with files above/below 5 MB threshold
- Verified page-count threshold with a 3 MB / 150-page text PDF correctly routes to `pdf_hybrid`
- `DocxHybridParser` tested with a real 987 KB Robustel product deck (pptx) — LibreOffice conversion + page renders succeeded
- `go build ./...` passes

### DIFF
--- a/docreader/parser/office_hybrid_parser.py
+++ b/docreader/parser/office_hybrid_parser.py
@@ -1,0 +1,140 @@
+"""
+Office format hybrid parsers: MarkItDown text extraction + LibreOffice page rendering.
+
+Both DocxHybridParser and PptxHybridParser follow the same pattern:
+  1. MarkitdownParser  — extracts text structure concurrently
+  2. LibreOffice       — converts file to PDF, then PDFScannedParser renders
+                         every page as a JPEG for the multimodal VLM pipeline
+
+The two passes run in a ThreadPoolExecutor so total wall-clock time is
+max(text_extraction, lo_conversion + page_render) rather than their sum.
+"""
+
+import concurrent.futures
+import logging
+import os
+import shutil
+import subprocess
+import tempfile
+
+from docreader.models.document import Document
+from docreader.parser.base_parser import BaseParser
+from docreader.parser.markitdown_parser import MarkitdownParser
+from docreader.parser.pdf_parser import PDFScannedParser
+
+logger = logging.getLogger(__name__)
+
+# LibreOffice search order: PATH first, then common installation paths.
+_SOFFICE_FALLBACKS = [
+    "/opt/homebrew/bin/soffice",
+    "/usr/bin/soffice",
+    "/usr/local/bin/soffice",
+    "/Applications/LibreOffice.app/Contents/MacOS/soffice",
+]
+
+SOFFICE_TIMEOUT = 120  # seconds per conversion
+
+
+def find_soffice() -> str | None:
+    """Return the path to the LibreOffice soffice binary, or None."""
+    path = shutil.which("soffice")
+    if path:
+        return path
+    for candidate in _SOFFICE_FALLBACKS:
+        if os.path.isfile(candidate):
+            return candidate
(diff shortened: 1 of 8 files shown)


toy0116:
Closing as largely superseded by upstream changes since this PR was opened:

- **Hybrid PDF parsing** — `PDFParser` was rewritten into a per-page router (native-text pages keep their text layer; scanned pages render to JPEG; both interleave in reading order), and a new `OpenDataLoaderParser` engine was added — together covering the image-rich PDF case this PR's `PDFHybridParser` targeted.
- **Queue isolation** — `main` now has dedicated `multimodal` / `graph` / `question` lanes plus configurable concurrency, addressing throughput from a different angle than this PR's `doc_large` size-routing.

Rebasing the 7-commit chain is no longer a mechanical merge. If a still-unique piece is worth it (e.g. ≥5 MB size-based queue routing as a small standalone change), I'll open a focused PR for just that. Thanks!

PR #1233, opened 2026-05-09
fix(agent): detect and retry planning-artifact responses

toy0116:
## Problem

The ReAct agent loop occasionally emits a short "planning statement" as its final answer — e.g. `"Now let me fetch the latest product specs…"` — when the LLM stops (`finish_reason=stop`) without actually calling any tools. This happens because the model decides to verbalize its intent before acting, then stops prematurely.

The user sees the intent sentence as the final answer instead of the actual retrieved information.

### Commit 1 — detect and retry

Adds `isPlanningArtifact(content string) bool` to `observe.go`:

- Heuristic: content ≤400 chars **and** starts with a known planning prefix (`"let me"`, `"i will now"`, `"i need to"`, etc.) **and** contains an action verb (`fetch`, `search`, `retrieve`, …)
- Also catches "escape to web" patterns (`"from the web"`, `"online search"`, …) regardless of prefix
- When detected, returns a `responseVerdict{planningArtifact: true}` — same retry path as empty content

Adds `planningArtifact bool` field to `responseVerdict` and the `planningArtifactRetries` counter in `engine.go`, parallel to the existing `emptyRetries` mechanism.

### Commit 2 — nudge prompt

When retrying a planning artifact, the nudge message instructs the model to synthesize from already-retrieved data rather than asking it to call more tools. Previously the nudge asked the model to continue, which could trigger another round of tool calls instead of producing the answer.

## Testing

- Verified planning artifact detection fires on `"Now let me search for the documents"` (≤400 chars, planning prefix, action verb)
- Verified it does NOT fire on real answers that happen to mention fetching in passing (length guard)
- Verified the retry counter increments and the nudge message is appended to context

### DIFF
--- a/internal/agent/engine.go
+++ b/internal/agent/engine.go
@@ -624,11 +624,13 @@ func (e *AgentEngine) runReActIteration(
 		if verdict.planningArtifact {
 			*emptyRetries++
 			if *emptyRetries <= maxEmptyResponseRetries {
-				logger.Warnf(ctx, "[Agent][Round-%d] Planning artifact — nudging to use tools (%d/%d)",
+				logger.Warnf(ctx, "[Agent][Round-%d] Planning artifact — nudging to synthesize (%d/%d)",
 					round, *emptyRetries, maxEmptyResponseRetries)
 				*messagesPtr = append(*messagesPtr, chat.Message{
-					Role:    "user",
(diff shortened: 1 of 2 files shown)


lyingbug:
我不太确定这种通过关键词来匹配的能真正起到作用

PR #1279, opened 2026-05-12
fix: surface model connection test errors

langcaiye:
## Summary
- surface backend failure reasons in the model connection test dialog
- redact and truncate displayed error details before showing them to users
- provide a clearer hint when an upstream model endpoint returns HTML instead of JSON

## Tests
- npm run build
- CGO_CXXFLAGS="-isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/c++/v1" go test ./internal/handler


langcaiye:
Closing this PR to keep the current contribution scope focused on the Anthropic provider follow-up. We may revisit model connection error UX separately later.

PR #1290, opened 2026-05-12
fix: localize agent skill options

ziwon:
## Description
Localizes the built-in Agent Skills displayed in the agent editor.

The backend returns Skill metadata directly from `SKILL.md`, so the skill cards could appear in Chinese even when the UI locale is Korean or English. This PR adds display-only localized names and descriptions for known preloaded skills while preserving the original skill identifiers used by `selected_skills` and backend execution.

It also keeps the Skills menu gated by sandbox availability and defaults the loading state to visible until the backend explicitly reports `skills_available: false`, avoiding a brief disappearance during dependency loading.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue
N/A

## Testing
- Ran `npm ci` in `frontend/`
- Ran `npm run build-only` in `frontend/`
- Verified the change is limited to display text: checkbox values still use the original Skill `name`

## Checklist
- [x] Self-reviewed the code

## Screenshots / Recordings
Not attached. This is a small text-only UI localization fix for the Agent Skills selection list.

### DIFF
--- a/frontend/src/views/agent/AgentEditorModal.vue
+++ b/frontend/src/views/agent/AgentEditorModal.vue
@@ -839,14 +839,14 @@
                       <div class="setting-control setting-control-full">
                         <t-checkbox-group v-model="formData.config.selected_skills" class="skills-checkbox-group">
                           <t-checkbox
-                            v-for="skill in skillOptions"
+                            v-for="skill in localizedSkillOptions"
                             :key="skill.name"
                             :value="skill.name"
                             class="skill-checkbox-item"
                           >
                             <div class="skill-item-content">
-                              <span class="skill-name">{{ skill.name }}</span>
-                              <span class="skill-desc">{{ skill.description }}</span>
+                              <span class="skill-name">{{ skill.displayName }}</span>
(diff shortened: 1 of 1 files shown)


lyingbug:
Thanks for the localization contribution. Closing this PR because the specific legacy UI and bundled skills it translates have been retired.

The patch adds a hard-coded translation map for the old preloaded skills. #3033 removed those host-preloaded skills, and the agent editor now uses the tenant skill catalog introduced in #2852. The [current skill picker](https://github.com/Tencent/WeKnora/blob/b23ae055feef678c72138298f4d97693c6e49920/frontend/src/views/agent/AgentEditorModal.vue#L1371) is driven by installed catalog entries rather than the old `skillOptions` list.

This is an obsolete-target closure, not a claim that every installed skill's metadata is translated. Any remaining localization issue in the current catalog UI should be handled in a focused follow-up.

PR #1380, opened 2026-05-18
fix(model): allow updating built-in (global) models

### LINKED ISSUE #1373: [Bug]: Cannot work with global models (build in models)
### Affected Component

Frontend UI

### Bug Description

The UI and also API do not support to create global models (build in models).
When updating the database to set a model's `is_builtin` attribute to `true`, the process fails because the `GetByID` method requires the model to belong to a specific tenant. I suggest updating the method to:
```
func (r *modelRepository) GetByID(ctx context.Context, tenantID uint64, id string) (*types.Model, error) {
	var m types.Model
	if err := r.db.WithContext(ctx).Where(
		"id = ? AND (tenant_id = ? OR is_builtin = true)", id, tenantID,
	).First(&m).Error; err != nil {
		if errors.Is(err, gorm.ErrRecordNotFound) {
			return nil, nil
		}
		return nil, err
	}
	return &m, nil
}
```

### Steps to Reproduce

No way to mark a model as public/global

### Expected Behavior

From the UI should allow us to mark a models as public/global and the backend API support to process it.

### Actual Behavior

No way to mark a model as public/global

### WeKnora Version

latest

### Deployment Method

Docker Compose

### Operating System

Windows 11

### Relevant Logs

```shell

```

### Confirmation

guangyang1206:
Fixes #1373.

Previously, the Update function had Where("id = ? AND tenant_id = ?", ...) which prevented updating built-in models (is_builtin=true) from other tenants.

Now allows updates if:
  1. The model belongs to the tenant, OR
  2. The model is a built-in model (is_builtin = true)

This enables tenants to customize/configure global models for their usage.
### Issue Reference
Fixes #1373

### Problem Description
The UI and API do not support creating/updating global models (built-in models). When trying to update a built-in model (`is_builtin=true`), the operation fails because the `Update` function in the repository only allows updates when both `id` and `tenant_id` match.

### Root Cause
In `internal/application/repository/model.go`, the `Update` function had:
```go
Where("id = ? AND tenant_id = ?", m.ID, m.TenantID)
```
This prevented updating built-in models that belong to other tenants.

### Solution
Modified the `Where` clause to allow updates if EITHER condition is met:
1. The model belongs to the current tenant (`tenant_id` matches), OR
2. The model is a built-in model (`is_builtin = true`)

New code:
```go
Where("id = ? AND (tenant_id = ? OR is_builtin = true)", m.ID, m.TenantID)
```

### Impact
- Tenants can now update/configure global (built-in) models for their usage
- No breaking changes - the check is relaxed, not removed
- Maintains security by still requiring the `id` to match

### Testing
- [ ] Unit tests pass
- [ ] Manual testing: Try to update a built-in model from a different tenant

### Checklist
- [x] Code follows project coding style
- [x] Commit message follows Conventional Commits
- [ ] Tests added/updated (if applicable)

### DIFF
--- a/internal/application/repository/model.go
+++ b/internal/application/repository/model.go
@@ -65,8 +65,11 @@ func (r *modelRepository) List(
 // Update updates a model
 func (r *modelRepository) Update(ctx context.Context, m *types.Model) error {
 	// Use Select to explicitly update all fields, including zero values like false
+	// Allow updates if:
+	//   1. The model belongs to the tenant (tenant_id match), OR  
+	//   2. The model is a built-in model (is_builtin = true)
 	return r.db.WithContext(ctx).Debug().Model(&types.Model{}).Where(
-		"id = ? AND tenant_id = ?", m.ID, m.TenantID,
+		"id = ? AND (tenant_id = ? OR is_builtin = true)", m.ID, m.TenantID,
 	).Select("*").Updates(m).Error
 }
 


lyingbug:
这会导致任意用户都能通过接口修改内置模型的信息？

Issue #993, opened 2026-04-20
Phase 2: Per-KB VectorStore Binding


ochanism:
## Context

This is Phase 2 of the multi-store vector DB architecture proposal (#913), following Phase 1 (#921) which introduced VectorStore as a first-class entity with a dual-map Registry.

Phase 2 binds each knowledge base to a specific VectorStore, enabling per-KB routing of indexing and retrieval while preserving 100% backward compatibility for KBs that remain unbound.

## Goals

1. **Per-KB VectorStore binding** — `knowledge_bases.vector_store_id` nullable reference. NULL preserves the existing env-store flow (RETRIEVE_DRIVER fallback)
2. **Immutable after creation** — enforced at the GORM layer via `<-:create` (blocks every ORM UPDATE path including `Select().Updates()`) plus a service-layer update DTO that omits the field. Phase 4 cross-store migration will be the only sanctioned write path, using a dedicated raw-SQL repository method
3. **Factory function refactoring** — collapse 25 repeated `NewCompositeRetrieveEngine` call sites into `CreateRetrieveEngineForKB(ctx, registry, vectorStoreID)` plus `CreateRetrieveEngineFromPayload(registry, engines, vectorStoreID)` for async task handlers
4. **Registry interface extension** — promote `GetByStoreID` from concrete-only (Phase 1) to `interfaces.RetrieveEngineRegistry`
5. **Multi-store fan-out search** — group KBs by their bound store, retrieve in parallel via `errgroup`, merge with engine-aware score normalization when different engine types are mixed
6. **Defensive checks** — embedding-model consistency across multi-KB search and CopyKnowledgeBase, cross-store CopyKB rejection (cross-store copy is Phase 4), VectorStore delete safety (transaction + `SELECT FOR UPDATE` + KB binding count)
7. **100% backward compatible** — all existing KBs have `vector_store_id = NULL` after migration and continue to use the tenant's effective engines

## PR Roadmap (5 PRs)

Each PR is designed to be **independently mergeable** without breaking existing functionality.

```
PR 1 (~400 lines) ──→ PR 2 (~800 lines) ──→ PR 3 (~400 lines) ──→ PR 4 (~350 lines)
KB model + migration   Factory + refactor    API + guards         Multi-store search
                                                 │
                                                 └──→ PR 5 (~300 lines)
                                                      KB binding UI
```

| PR | Scope | Merge Safety |
|----|-------|--------------|
[...]

Issue #913, opened 2026-04-04
[Proposal] Multi-Store Vector DB Architecture, Per-KB Index Isolation & OpenSearch Support


ochanism:
### Scalability Limitations of Current Architecture

WeKnora's retrieval engine architecture works well for small-scale, single-backend environments, but structural limitations emerge when scaling to production.

**Limitation 1: Registry is a Singleton per EngineType**

The current `RetrieveEngineRegistry` is implemented as `map[RetrieverEngineType]RetrieveEngineService`. Attempting to register more than one engine of the same type results in a `"repository type %s already registered"` error.

```go
// registry.go — current structure
type RetrieveEngineRegistry struct {
    repositories map[types.RetrieverEngineType]interfaces.RetrieveEngineService
}

func (r *RetrieveEngineRegistry) Register(repo interfaces.RetrieveEngineService) error {
    if _, exists := r.repositories[repo.EngineType()]; exists {
        return fmt.Errorf("repository type %s already registered", repo.EngineType())
    }
    // ...
}
```

This makes it impossible to run multiple instances of the same DB type. For example, you cannot operate two ES clusters for hot/warm tiers or regional separation. (Different DB types — e.g., one ES + one Qdrant — can coexist since they have different EngineTypes, but two instances of the same type cannot.)

**Limitation 2: RETRIEVE_DRIVER is a Global Environment Variable**

`initRetrieveEngineRegistry()` in `container.go` parses `os.Getenv("RETRIEVE_DRIVER")` once at startup, creating a single client. All Knowledge Bases are forced to share the same vector DB instance.

```go
// container.go — current initialization flow
func initRetrieveEngineRegistry(db *gorm.DB, cfg *config.Config) {
    retrieveDriver := strings.Split(os.Getenv("RETRIEVE_DRIVER"), ",")
    // "elasticsearch_v8" → creates exactly one ES client
    // all KBs share this instance
}
```

**Limitation 3: All KBs Share a Single Index/Collection**

Even within the same vector DB instance, all KB data is stored in **a single index (or collection)**. KB isolation relies solely on filtering by the `knowledge_base_id` field within documents.

```go
// elasticsearch/v8/repository.go — all KB data in a single index
indexName := os.Getenv("ELASTICSEARCH_INDEX")  // default: "xwrag_default"
res := &elasticsearchRepository{client: client, index: indexName}
// → all KB vectors go into this single index

// filtering by knowledge_base_id at query time
[...]

PR #1427, opened 2026-05-21
feat(frontend): expose KB ↔ vector store binding in editor, list, and detail

ochanism:
## Summary

Wire the Phase 2 backend (#994 / #1310 / #1372 / #1386) into the KB editor, list, and detail screens so users can choose which vector store a knowledge base is bound to and see the binding wherever a KB is displayed. The dropdown is opt-in — existing KBs and clients that don't select a store continue to use the env-store fallback unchanged.

In addition to the UI surface, the `ListKnowledgeBases` response is now enriched with the same `vector_store_*` view fields the per-KB endpoints already return (PR #1372), through a single batched store lookup so the list endpoint stays at one round trip regardless of how many KBs the tenant owns.

Cross-organization shared KBs strip `vector_store_id` / `vector_store_name` / `vector_store_engine_type` end-to-end — the share-list endpoints now go through the same projection helper as the single-KB endpoints, so the owning tenant's store inventory never leaks across organizations.

## Context

Part of [#993](https://github.com/Tencent/WeKnora/issues/993) — Phase 2 (Per-KB VectorStore Binding) of the [multi-store roadmap](https://github.com/Tencent/WeKnora/issues/913).

Phase 2 PR sequence (this is PR 5 of 5, the final one):

| # | PR | Status |
|---|----|--------|
| 1 | [#994](https://github.com/Tencent/WeKnora/pull/994) — `vector_store_id` column + migration | MERGED |
| 2 | [#1310](https://github.com/Tencent/WeKnora/pull/1310) — KB-scoped retrieve-engine factory + 25-site refactor | MERGED |
| 3 | [#1372](https://github.com/Tencent/WeKnora/pull/1372) — KB create/retrieve API + defensive logic | MERGED |
| 4 | [#1386](https://github.com/Tencent/WeKnora/pull/1386) — multi-store fan-out search | MERGED |
| **5** | **this PR — KB binding UI + list-response enrich** | **here** |

Depends on #994, #1310, #1372, and #1386 — all four are merged into `main`.

## Changes at a glance

- KB editor modal gains a new "Vector Store" navigation section. Create mode shows a dropdown that lists the system default (env-store) and every tenant-owned `VectorStore` (`source === 'user'`), each labeled with its underlying engine type. Edit mode hides the dropdown and shows the bound store read-only with an explicit immutability hint — the backend's `UpdateKnowledgeBase` request DTO has never accepted `vector_store_id`, and the UI now mirrors that contract.
- KB list cards show an engine-type badge (postgres / qdrant / …) for every KB whose store is currently reachable, regardless of whether the binding is env-default or user-owned. A red `unavailable` variant is shown for KBs whose backing store is currently unreachable (deleted store row or registry miss).
- KB detail header shows the bound store via the new `VectorStoreBadge` component. Cross-organization shared KBs fall through to a single "Shared" badge with no store metadata, matching the backend's stripped response.
- `ListKnowledgeBases` response is enriched with the same five `vector_store_*` fields the single-KB endpoints return, through `buildKBListResponse` + `batchResolveKBStoreViews`. The batched helper collects every distinct own-tenant store id in the page, issues a single `BatchResolveStoreView`, and reuses one pre-computed env-default view for the env-bound rows — so the list endpoint stays at one store-resolution round trip regardless of page size.
- Cross-organization shared KBs are now projected through the same `buildKBResponse` helper (with `SharedStoreDisplay()`) on both share-list endpoints (`ListSharedKnowledgeBases` and `ListOrganizationSharedKnowledgeBases`), so `vector_store_id` / `vector_store_name` / `vector_store_engine_type` are stripped before the response leaves the server. The owning tenant's store inventory is not visible to a peer organization.
- Create-time typed errors (2200 / 2201 from #1372) are mapped to user-friendly messages and the editor jumps to the vector-store section so the user can pick a different store without re-navigating.

### Production — backend

| File | Change |
|------|--------|
| `internal/types/interfaces/vectorstore.go` | Adds `EnvDefaultStoreView(ctx) types.StoreDisplay` to the `VectorStoreService` interface so list-path callers can render env-bound rows with the same engine label as user-bound rows, without violating the N+1 invariant by issuing one `ResolveStoreView` per env-bound KB |
| `internal/application/service/vectorstore.go` | Implements `EnvDefaultStoreView` plus a shared `defaultStoreDisplay()` helper. Both `ResolveStoreView` and `BatchResolveStoreView` env paths now route through the same helper so the env-store label and engine type are computed in one place. The engine type is read from the cached `envStores[0].EngineType`, so no extra registry lookup is added per call |
| `internal/handler/knowledgebase.go` | New helpers: `buildKBListResponse(ctx, kbs, callerTenantID)` projects a page of KBs through `buildKBResponse` reusing a pre-computed env-default view; `batchResolveKBStoreViews(ctx, kbs, callerTenantID)` collects distinct own-tenant store ids and resolves them in one `BatchResolveStoreView` call (foreign-tenant and env-bound KBs are partitioned out so the batch query stays sound); `envDefaultStoreView(ctx)` is the thin handler-side adapter over the service method; `sharedKBRow(info, extras)` projects a `SharedKnowledgeBaseInfo` through `buildKBResponse` with `SharedStoreDisplay()` so cross-tenant strip applies consistently. Both list call sites (the agent-filter branch and the default branch) now route through `buildKBListResponse` |
| `internal/handler/organization.go` | `ListSharedKnowledgeBases` and `ListOrganizationSharedKnowledgeBases` build their response rows through `sharedKBRow` so every cross-organization shared KB has the store-identifying fields stripped before the response leaves the server. `is_mine` / `source_from_agent` are passed through as `extras` to preserve the existing space-view UI |

### Production — frontend

| File | Change |
|------|--------|
| `frontend/src/api/knowledge-base/index.ts` | Extends the `KnowledgeBaseStoreView` interface (new) with the five read-only `vector_store_*` fields populated by the list / get / create / update / pin responses, plus the new `VectorStoreSource = 'env' \| 'user' \| 'shared' \| 'unavailable'` and `VectorStoreStatus = 'available' \| 'unavailable'` enums. `createKnowledgeBase` also accepts an opt-in `vector_store_id` — `updateKnowledgeBase` intentionally does not, matching the backend DTO |
| `frontend/src/components/VectorStoreBadge.vue` (new) | Single-source badge component covering the four `source` values. `env` and `user` render the engine type in the brand-color tint; `shared` renders a neutral chip with no engine label; `unavailable` renders an error-color tint with a clear warning. Used by both the list cards and the detail header so the visual treatment stays consistent across surfaces |
| `frontend/src/views/knowledge/KnowledgeBaseEditorModal.vue` | Adds a "Vector Store" navigation tab to the editor's left sidebar (between "Basic info" and "Knowledge", matching the placement of other infrastructure settings). The section body is delegated to `KBVectorStoreSettings`. Create-time typed AppErrors (2200 `vector store not found` / 2201 `vector store unavailable`) are caught and mapped to friendly messages, and the editor auto-navigates to this section so the user lands on the right control without manually clicking through |
| `frontend/src/views/knowledge/settings/KBVectorStoreSettings.vue` (new) | The section body. Create mode renders a `t-select` dropdown with the system default and every `source === 'user'` store (each labeled with its engine type) plus a "Go to Vector Store Settings" link that opens the global settings panel via `uiStore.openSettings('vectorstore')` — avoiding the modal-stack breakage that a `router.push` would cause. Edit mode renders a read-only `VectorStoreBadge` plus an immutability hint |
| `frontend/src/views/knowledge/KnowledgeBaseList.vue` | Both card templates (the main list driven by `filteredKnowledgeBases` and the `mine` tab driven by `sortedMineKbs`) render a store badge for KBs whose binding is currently reachable, plus the warning variant when `vector_store_status === 'unavailable'`. The badge HTML is shared between the two templates and a small amount of CSS is added to align with the existing `feature-badge` palette (brand-color tint for available, error-color for unavailable) |
| `frontend/src/views/knowledge/KnowledgeBase.vue` | The detail-page header renders the same `VectorStoreBadge` so the single-KB view matches the list-view affordances |
| `frontend/src/i18n/locales/{en-US,ko-KR,zh-CN}.ts` | 18 new keys covering the editor section title and helper text, dropdown labels, immutability hint, "Go to Settings" link, badge tooltips, and the typed-error friendly messages. Full translations for en / ko / zh |
| `frontend/src/i18n/locales/ru-RU.ts` | Same 18 keys with the English text as placeholders, pending a Russian translation pass |

### Tests

| File | Change |
|------|--------|
| `internal/handler/knowledgebase_pr5_list_test.go` (new) | Covers `buildKBListResponse` across the three source variants (env-bound / user-bound / shared) and pins the two invariants the helper is responsible for: (1) batched store resolution — `TestListKB_BatchesStoreLookupsToAvoidNPlus1` asserts that resolving a page with N user-bound KBs across two distinct stores issues exactly one `BatchResolveStoreView` call (not one per row); (2) graceful degradation — `TestListKB_GracefullyDegradesWhenStoreResolverFails` asserts that a `BatchResolveStoreView` failure falls back to per-row `ResolveStoreView` so a transient batch failure does not blank the page. Additional cases cover the env-bound row using the pre-computed env default, the shared row going through `SharedStoreDisplay()`, and the foreign-tenant strip being applied even when the caller's tenant id matches by accident |

## Backward compatibility

| Surface | Compatibility analysis |
|---------|------------------------|
| KB list page | Existing tenants see exactly the same set of KBs as before. Every card gains an engine-type badge (postgres for env-bound today, qdrant for KBs explicitly bound to a Qdrant store, etc.). Tenants that have never opened the new editor section still get the badge because the backend now enriches the list response from the env-store config |
| KB editor — open existing KB | Edit mode hides the dropdown and shows the bound store read-only. No behavior change to fields the user could edit before |
| KB editor — create new KB | The dropdown is opt-in. Leaving "System default" selected omits `vector_store_id` from the create request, which is the same payload the editor produced before this PR |
| `ListKnowledgeBases` response | Adds five new optional fields (`vector_store_id`, `vector_store_name`, `vector_store_engine_type`, `vector_store_source`, `vector_store_status`) to each row, mirroring the per-KB endpoint shape from #1372. Existing clients that ignore unknown fields are unaffected |
| `ListSharedKnowledgeBases` / `ListOrganizationSharedKnowledgeBases` | Now strip `vector_store_id` / `vector_store_name` / `vector_store_engine_type` from cross-organization rows. Behaviorally tightens the existing leak surface — clients consuming those fields from a shared KB row were depending on undocumented behavior. `vector_store_source = 'shared'` is set so clients can distinguish stripped rows from a missing binding |
| Store resolution N+1 | `ListKnowledgeBases` previously made zero store-resolution calls (it returned the GORM row directly). It now makes exactly one `BatchResolveStoreView` call per page plus one `EnvDefaultStoreView` lookup that is in-memory after the first call. The list endpoint stays at one store-side round trip regardless of page size |
| Frontend i18n | Three locales (en / ko / zh) get full translations. Russian receives the English text as a placeholder so the UI does not show empty strings if a tenant has the Russian locale active; a follow-up will replace those with real translations |
| Migrations | None |

## Behavior change (intentional)

- **Cross-organization shared KBs no longer expose `vector_store_id` / `vector_store_name` / `vector_store_engine_type`.** Prior to this PR these fields could appear on shared rows because the share-list endpoints serialized the raw `KnowledgeBase` struct. They are now stripped via `sharedKBRow` so the owning tenant's store inventory is not visible to a peer organization. Clients that need to distinguish a shared row from an env-bound row should read `vector_store_source === 'shared'`.
- **`ListKnowledgeBases` rows now include `vector_store_*` enrichment.** Previously only the single-KB endpoints (`get` / `create` / `update` / `pin`) returned those fields. The list endpoint now matches.
- **Edit mode in the KB editor exposes the binding as read-only.** The backend already rejected attempts to modify `vector_store_id` after creation; the UI now makes that contract visible by hiding the control rather than disabling it.

## Known limitations

- **`ListOrganizationSharedKnowledgeBases` strips even own-tenant rows.** When the caller's own KB happens to be in the listed organization's space view, that row goes through the same `sharedKBRow` projection as any other row and therefore loses its store metadata. The standard `ListKnowledgeBases` and per-KB endpoints continue to enrich own-tenant rows correctly, so this only affects the space view. Allowing the space view to enrich own-tenant rows specifically requires `OrganizationHandler` to gain a `vectorStoreService` dependency and is deferred to a follow-up cleanup PR.
- **`vector_store_status: "unavailable"` is not a live health probe.** The backend sets this signal when the bound store record has been soft-deleted, the registry has no entry for the id, or the resolution path hit a transient infrastructure error. It is not driven by an active health check against the store, so a network blip that the registry has not yet observed will still show `status: 'available'`. The frontend renders a warning badge consistently for whichever path produced the signal.
- **Typed errors 2200 / 2201 are hard to trigger from the editor itself.** The dropdown only lists stores the backend currently reports as available, so picking a "broken" id requires going around the UI (e.g. with a direct cURL call). The handling exists as a defense-in-depth path; the user-facing flow is dominated by the happy path.
- **`ru-RU` keys are English placeholders.** Three of the four locales get full translations in this PR; Russian follows in a separate localisation pass.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -count=1 -race ./internal/handler/...` — including the new `knowledgebase_pr5_list_test.go` (env / user / shared variants + the two N+1 / graceful-degradation invariants)
- [x] `go test -count=1 -race ./internal/application/service/...` (`EnvDefaultStoreView` + `defaultStoreDisplay` helper coverage; one pre-existing OSS-mock failure `TestOssEnsureBucket_CreateFails` is reproducible on `main` and unrelated to this PR)
- [x] `go test -count=1 -race ./internal/application/service/retriever/... ./internal/types/...`
- [x] `pnpm build-only` (Vite production build, type-check parity with `main` — this PR adds no new TypeScript errors)
- [x] Local manual smoke against `docker compose dev` (PG env + Qdrant + DocReader + MinIO + Redis), two tenants with the cross-organization share path exercised:
  - **List enrich**: list response carries `vector_store_*` for every row, env-bound rows resolve to "System default / postgres", user-bound rows resolve to the bound `VectorStore` name + engine
  - **Editor — create**: dropdown lists "System default (postgres)" + every user-owned store labeled with its engine; selecting Qdrant produces a KB whose `vector_store_id` is set in the DB
  - **Editor — edit**: dropdown disappears, read-only badge + immutability hint render; a manual PUT with `vector_store_id` is silently ignored (backend DTO rejects the field, other fields update normally)
  - **Cross-organization share**: a tenant sharing two KBs (one env-bound, one user-bound) to a peer org → the peer's `/shared-knowledge-bases` response has all three `vector_store_id` / `vector_store_name` / `vector_store_engine_type` keys absent and only `vector_store_source: 'shared'` set
  - **Store unavailable**: soft-deleting a store row → the bound KB's row shows `vector_store_status: 'unavailable'` and the badge renders the error variant; other KBs are unaffected
  - **Typed AppError**: cURL POST with an invalid `vector_store_id` (zero UUID or foreign-tenant UUID) → HTTP 400 with `{"error":{"code":2200,"message":"vector store not found"}}`; the existence of a foreign-tenant store is not revealed
  - **i18n**: en / ko / zh render the 18 new keys; ru shows the English placeholders pending the localisation pass

### DIFF
--- a/frontend/src/api/knowledge-base/index.ts
+++ b/frontend/src/api/knowledge-base/index.ts
@@ -20,6 +20,38 @@ export function listKnowledgeBases(params?: {
   return get(qs ? `/api/v1/knowledge-bases?${qs}` : '/api/v1/knowledge-bases');
 }
 
+// Read-only vector-store binding metadata enriched onto every KB
+// response (list, create, get, update, pin). Source carries where the
+// binding points; status reports whether that target is currently
+// reachable by the server.
+//
+//   - source 'env'    → KB uses the tenant's env-configured store
+//                       (RETRIEVE_DRIVER). vector_store_id is null and
+//                       vector_store_name is the localized "System
+//                       default" label; vector_store_engine_type still
+//                       reports the underlying engine (e.g. "postgres").
+//   - source 'user'   → KB is bound to a tenant-owned VectorStore.
+//                       vector_store_id / name / engine_type are real.
+//   - source 'shared' → KB belongs to a different tenant and is
+//                       readable via cross-organization sharing. The
+//                       server strips vector_store_id and engine_type
+//                       to avoid leaking the owner tenant's store
+//                       inventory; only this source marker arrives.
+//   - status 'unavailable' → the binding cannot be reached right now
+//                       (deleted row, registry miss, transient infra
+//                       failure). Operators recover via the global
+//                       Vector Stores settings page.
+export type VectorStoreSource = 'env' | 'user' | 'shared' | 'unavailable';
+export type VectorStoreStatus = 'available' | 'unavailable';
+
+export interface KnowledgeBaseStoreView {
+  vector_store_id?: string | null;
+  vector_store_name?: string;
+  vector_store_engine_type?: string;
+  vector_store_source?: VectorStoreSource;
+  vector_store_status?: VectorStoreStatus;
+}
+
 export function createKnowledgeBase(data: {
   name: string;
   description?: string;
@@ -27,6 +59,11 @@ export function createKnowledgeBase(data: {
   chunking_config?: any;
   embedding_model_id?: string;
   summary_model_id?: string;
+  // Opt-in binding to a specific tenant-owned VectorStore. Omit (or
+  // send undefined / empty string) to fall back to the env-configured
+  // store. Immutable after creation — UpdateKnowledgeBase intentionally
+  // does not accept this field.
+  vector_store_id?: string;
   vlm_config?: {
     enabled: boolean;
     model_id?: string;
--- a/frontend/src/components/VectorStoreBadge.vue
+++ b/frontend/src/components/VectorStoreBadge.vue
@@ -1,0 +1,114 @@
+<template>
+  <span :class="['vs-badge', `vs-badge-${effectiveSource}`, isUnavailable && 'vs-badge-warn']">
+    <t-icon :name="iconName" class="vs-badge-icon" />
+    <span class="vs-badge-name">{{ displayName }}</span>
+    <span
+      v-if="engineType && (effectiveSource === 'user' || effectiveSource === 'env')"
+      class="vs-badge-engine"
+    >
+      ({{ engineType }})
+    </span>
+    <t-tag
+      v-if="isUnavailable"
+      theme="danger"
+      variant="light"
+      size="small"
+      class="vs-badge-warn-tag"
+    >
+      {{ $t('vectorStoreBadge.unavailable') }}
+    </t-tag>
+  </span>
+</template>
+
+<script setup lang="ts">
+import { computed } from 'vue'
+import { useI18n } from 'vue-i18n'
+import type { VectorStoreSource, VectorStoreStatus } from '@/api/knowledge-base'
+
+const props = defineProps<{
+  source?: VectorStoreSource
+  name?: string
+  engineType?: string
+  status?: VectorStoreStatus
+}>()
+
+const { t } = useI18n()
+
+// When backend omits the source (e.g. legacy KB row from a cached list
+// endpoint that does not enrich), treat it as env so the badge renders
+// gracefully instead of going blank.
+const effectiveSource = computed<VectorStoreSource>(() => props.source || 'env')
+
+const isUnavailable = computed(
+  () => props.status === 'unavailable' || effectiveSource.value === 'unavailable',
+)
+
+const iconName = computed(() => {
+  switch (effectiveSource.value) {
+    case 'env':
+    case 'user':
+      // Both env- and user-bound KBs sit on top of a vector store; the
+      // distinction is purely organizational (configured at process
+      // start vs. created in the UI), so they share the same icon.
+      return 'data-base'
+    case 'shared':
+      return 'share'
+    case 'unavailable':
+    default:
+      return 'help-circle'
+  }
+})
+
+const displayName = computed(() => {
+  if (effectiveSource.value === 'env') return t('vectorStoreBadge.systemDefault')
+  if (effectiveSource.value === 'shared') return t('vectorStoreBadge.sharedFromOrg')
+  return props.name || t('vectorStoreBadge.unknownStore')
+})
+</script>
+
+<style scoped>
+.vs-badge {
+  display: inline-flex;
+  align-items: center;
+  gap: 4px;
+  padding: 2px 8px;
+  border-radius: 4px;
+  font-size: 12px;
+  line-height: 1.4;
+  background: var(--td-bg-color-component, #f5f7fa);
+  color: var(--td-text-color-primary, #1d2129);
+}
+
+.vs-badge-env {
+  background: var(--td-brand-color-1, #ecf2fe);
+  color: var(--td-brand-color-7, #0052d9);
+}
+
+.vs-badge-user {
+  background: var(--td-success-color-1, #e8f8f2);
+  color: var(--td-success-color-7, #00754a);
+}
+
+.vs-badge-shared {
+  background: var(--td-warning-color-1, #fff1e9);
+  color: var(--td-warning-color-7, #b85b00);
+}
+
+.vs-badge-warn {
+  background: var(--td-error-color-1, #fde9e6);
+  color: var(--td-error-color-7, #b32700);
+}
+
+.vs-badge-icon {
+  font-size: 14px;
+}
+
+.vs-badge-engine {
+  opacity: 0.7;
+  font-size: 11px;
+}
+
+.vs-badge-warn-tag {
+  margin-left: 4px;
+}
+</style>
--- a/frontend/src/i18n/locales/en-US.ts
+++ b/frontend/src/i18n/locales/en-US.ts
@@ -1708,6 +1708,7 @@ export default {
     sidebar: {
       basic: 'Basic Information',
       models: 'Model Configuration',
+      vectorStore: 'Vector Store',
       chunking: 'Chunking Settings',
       storage: 'Storage Engine',
       advanced: 'Advanced Settings',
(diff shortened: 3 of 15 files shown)


ochanism:
image

image

image

PR #1483, opened 2026-05-26
feat(builtin-models): add YAML lifecycle (delete, resurrect, default invariant) on top of #1453

lyingbug:
## Summary

Builds on #1453 to close the lifecycle gaps that the original PR
intentionally left open. With this change `config/builtin_models.yaml`
becomes a complete source of truth for the slice of `models` it owns:
edit, add, **and remove** all happen by editing the file.

The first commit on this branch is #1453 verbatim (so the diff is
self-contained while #1453 is still open); the three commits added here
are the lifecycle work.

### What's new on top of #1453

**1. `models.managed_by` column** (commit `1d2685f2`)
- Tags every YAML-written row with `managed_by='yaml'`. Rows created via
  UI/API/SQL keep the default empty string and are invisible to the
  loader — manual-management workflows are not touched.
- PG: new migration `000052_models_managed_by` adds the column plus a
  partial index over non-empty values to keep the startup reconcile
  query cheap.
- SQLite: `migrations/sqlite/000000_init.up.sql` is updated in place
  per Lite-mode convention.

**2. YAML lifecycle reconcile** (commit `e482aa28`)
- UPSERT now includes `deleted_at` in `DoUpdates`, so a YAML-managed row
  that was previously soft-deleted (e.g. via UI/API) is **resurrected**
  when it reappears in the file. Closes the "ghost row that exists in
  the table but is invisible" failure mode.
- After all UPSERTs, the loader **soft-deletes** rows where
  `managed_by='yaml' AND id NOT IN <current YAML id set>`. Removing an
  entry from YAML is now the supported way to retire a built-in model —
  no manual SQL needed.
- When a YAML entry sets `is_default: true`, the loader first clears
  `is_default` on other rows in the same `(tenant_id, type)` bucket,
  mirroring the invariant enforced by the API path
  (`repository.UnsetDefaultModel`). Prevents multiple defaults from
  accumulating when ops mix YAML and UI.
- Defensive failure handling preserved: parse error / file missing /
  per-entry UPSERT failure all skip the sweep so a malformed file
  cannot mass-delete rows.

**3. Compose compatibility fix** (commit `c1ff3d4c`)
- #1453 introduced the map form of `env_file:` (`- path: .env, required: false`)
  which only works on Docker Compose v2.24+ (Jan 2024). Older deploys
  fail to parse `docker-compose.yml` even when `builtin_models.yaml` is
  not used — breaking the "no impact unless opted in" promise.
- Switched to the universally-supported array form `env_file: [.env]`.
  The trade-off (Compose errors when `.env` is absent) is covered by
  the existing `check_env_file` in `scripts/start_all.sh` and by a new
  guard added to `make docker-run` / `make docker-restart`.

### Tests

`internal/types/builtin_models_config_test.go` covers 11 scenarios
against an in-memory SQLite DB:

- File missing — no-op, no sweep, pre-existing yaml-managed rows preserved
- YAML parse error — warn + return, sweep skipped (regression guard for
  the "one typo wipes everything" failure mode)
- Basic UPSERT with defaults (tenant_id=10000, source=remote, status=active)
- Idempotent reload (no duplicates after second call)
- `${ENV}` interpolation: set value substituted, unset left as literal
- Drift sweep removes YAML rows that disappeared from the file
- Drift sweep ignores rows with `managed_by=''` (manual SQL seeds)
- Soft-deleted row resurrected when it reappears in YAML
- `is_default: true` clears existing default in same (tenant_id, type)
- Explicit `builtin_models: []` sweeps all yaml-managed rows
- Regression guard: BeforeCreate must not overwrite YAML-supplied stable ids

`go build ./...`, `go vet ./internal/types/...`, and
`go test ./internal/types/...` all pass.

### Operator-facing behavior

| Action | Before this PR | After |
|---|---|---|
| Add a model | edit YAML, restart | (unchanged) |
| Update a model | edit YAML, restart | (unchanged) |
| **Remove a model** | edit YAML, then manually `UPDATE`/`DELETE` in DB | **edit YAML, restart — done** |
| Resurrect a soft-deleted YAML model | manually `UPDATE models SET deleted_at=NULL` | re-add to YAML, restart |
| Promote a model to default via YAML | could leave multiple defaults if user had set one via UI | YAML now clears prior defaults in same bucket |
| Old Docker Compose (< v2.24) | `docker-compose.yml` failed to parse | works |

### Migration / rollout

- The schema migration adds a column with a safe default (`''`); no
  backfill needed since existing rows are exactly the manual-managed
  rows the loader is meant to leave alone.
- Down migration drops the column cleanly.
- SQLite uses the standard in-place init pattern; fresh Lite databases
  pick up the column from the existing `000000_init` migration.

## Type of Change

- [x] ✨ New feature (lifecycle on top of #1453)
- [x] 🐛 Bug fix (Compose v2.24+ requirement leaked into base config)
- [x] 🧪 Tests

## Testing

- `go build ./...` — passes
- `go vet ./internal/types/...` — clean
- `go test ./internal/types/...` — 11/11 new tests pass + existing tests
- Manual: simulated the "remove entry → restart" flow against a local
  SQLite-backed Lite instance; row is soft-deleted, re-adding the entry
  resurrects it with `deleted_at=NULL`.

## Notes for reviewers

- This PR depends on #1453's first commit. Two paths to land:
  1. Merge #1453 first, then rebase this branch onto main (drops the
     duplicate commit, leaves three clean commits).
  2. Close #1453 in favor of this branch.
  Happy with either — let me know.
- Original #1453 review comments about `varchar(36)` primary key length
  and `fmt.Printf` vs structured logger are intentionally **not**
  addressed here to keep the diff focused on lifecycle. Those are easy
  follow-ups.

## Checklist

- [x] Self-reviewed the code
- [x] Updated documentation (`docs/BUILTIN_MODELS.md`)
- [x] Breaking changes called out: none for end users; ops who relied on
      "YAML never touches the DB after first insert" will see soft
      deletes of yaml-managed rows that disappear from the file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,7 +123,12 @@ docker-build-frontend:
 docker-build-all: docker-build-app docker-build-docreader docker-build-frontend
 
 # Run Docker container (传统方式)
+# Touch .env if missing — docker-compose.yml's `env_file: [.env]` is required
+# for ${ENV} interpolation in builtin_models.yaml and would otherwise refuse
+# to parse on fresh clones. `start-all` handles this via check_env_file; this
+# direct path needs its own guard.
 docker-run:
+	@[ -f .env ] || ([ -f .env.example ] && cp .env.example .env || touch .env)
 	docker-compose up
 
 # 使用新脚本启动所有服务
@@ -164,6 +169,7 @@ clean-images:
 
 # Restart Docker container (stop, start)
 docker-restart:
+	@[ -f .env ] || ([ -f .env.example ] && cp .env.example .env || touch .env)
 	docker-compose stop -t 60
 	docker-compose up
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,13 +51,18 @@ services:
       timeout: 10s
       retries: 3
       start_period: 60s
-    # Optional: source .env so deployment-specific vars (e.g. those referenced
-    # by config/builtin_models.yaml via ${ENV}) reach the app container without
-    # having to be listed individually under `environment:`. Falls back
-    # silently when .env is absent (e.g. fresh upstream clone).
+    # Source .env so deployment-specific variables (e.g. those referenced by
+    # config/builtin_models.yaml via ${ENV}, or any future feature) reach the
+    # app container without having to be enumerated under `environment:`.
+    #
+    # NOTE: the array form below is used instead of the map form
+    # (env_file: - path: .env, required: false) because the map syntax
+    # requires Docker Compose v2.24+ (Jan 2024) and would break parsing on
+    # older deploys. The trade-off is that Compose errors out if .env is
+    # absent — see scripts/start_all.sh for the pre-flight touch that keeps
+    # fresh clones working.
     env_file:
-      - path: .env
-        required: false
+      - .env
     environment:
       - LOG_LEVEL=${LOG_LEVEL:-}
       - COS_SECRET_ID=${COS_SECRET_ID:-}
--- a/docs/BUILTIN_MODELS.md
+++ b/docs/BUILTIN_MODELS.md
@@ -237,7 +237,18 @@ WHERE id = '模型ID';
 
 ## 移除内置模型
 
-YAML 配置删除条目后**不会**自动清理 DB 里的对应行（避免误删 ops 手工维护的内置模型）。如需删除：
+**从 YAML 删除条目即可。** 应用启动时会自动软删除 `models` 表中 YAML 不再声明的 YAML 托管行 —— 你不再需要手工跑 SQL。
+
+工作原理：每条由 YAML 写入的行会被打上 `managed_by = 'yaml'` 标记。重启时 loader 走两步：
+
+1. UPSERT 当前 YAML 中的所有条目（按 `id` 幂等，包含把之前软删过的 `deleted_at` 重置为 NULL —— 也就是说从 YAML 拿掉再加回来等于"复活"）
+2. 软删除 `is_builtin = true AND managed_by = 'yaml' AND id NOT IN (当前 YAML 中的 id 集合)` 的行
+
+**手工通过 SQL 插入的 builtin 行（`managed_by = ''`）永远不会被 loader 触碰**，与 YAML 完全隔离。
+
+### 手工路径补充
+
+如果你是走 SQL 路径管理的（`managed_by = ''`），删除仍然走老方法：
 
 ```sql
 -- 取消 builtin 标记，恢复为普通模型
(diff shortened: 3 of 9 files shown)


lyingbug:
Pushed the two follow-ups that the PR body marked as deferred, plus two related polishes from the original review thread.

**New commits**

- `cbadfeba refactor(builtin-models): unify logging via stdlib log with stable prefix`
  Replaces `fmt.Printf` with `log.Printf("[builtin-models] ...")`. The natural choice (`logger.Infof/Warnf`) would create an import cycle because `internal/logger` imports `internal/types`; this is the same constraint that already forces `model.go`'s crypto error logs onto stdlib log. The `[builtin-models]` prefix mirrors the `[crypto]` convention so grep/Splunk still works.

- `cab02d3c feat(builtin-models): validate YAML entries and align ID length with schema`
  Three correctness items in one commit:
  1. **ID length / struct tag drift** — `models.id` is `varchar(64)` on both PG and SQLite per the init migrations, but `Model.ID`'s tag said `varchar(36)`. Tag now matches; new `ModelIDMaxLen=64` constant; loader rejects too-long ids up front with a clear message instead of letting `INSERT` explode with a generic "value too long for type" error.
  2. **Field validation** — `Type`, `Status` now checked against the known enum sets (`KnowledgeQA / Embedding / Rerank / VLLM / ASR` for type, `active / downloading / download_failed / empty` for status). Misspellings like `type: knowledgeqa` (lowercase) or `type: LLM` were previously persisted as-is and produced rows that looked fine in the table but failed at provider-factory lookup time. `Source` is intentionally not validated since the provider matrix in `internal/models/*` keeps growing. Invalid entries are warned + skipped (not aborting), and excluded from the drift-sweep keep-set so a typo'd retry doesn't delete existing matching rows.
  3. **Magic number cleanup** — `DefaultBuiltinModelTenantID = 10000` replaces the hard-coded literal in `toModel()`. The invariant lives in three places (PG migration, SQLite migration, this constant); naming it makes the cross-reference grep-able.

**New tests**

`TestLoadBuiltinModelsConfig_RejectsInvalidEntries` adds five sub-cases (id-too-long, missing-type, lowercase-type, unknown-type, unknown-status). All 11 existing tests still pass; `go build ./...`, `go vet ./internal/types/...`, `go test ./internal/types/...` all clean.

**Still deferred**

Two larger items from the review thread remain out of scope here because they need bigger structural changes:

- `${ENV}` interpolation at YAML-text level can break parsing if env values contain `:` / `#` / quotes / newlines. A post-parse, per-string-field implementation would be safer but is a bigger refactor (would also affect `internal/config/config.go` which uses the same pattern).
- AES-GCM nonce randomization causes `parameters` to be rewritten with a different ciphertext each restart, which bumps `updated_at`. Skipping the UPSERT when plaintext is unchanged needs a "select-then-compare" round-trip and is best done together with the deterministic-encryption question.

Happy to spin these out as separate PRs if you want them.

PR #1509, opened 2026-05-27
feat(models): introduce WikiSynthesis as a first-class model type

jackson-jia-914:
## Summary

Wiki synthesis currently shares the `KnowledgeQA` model pool with the chat
LLM. The KB editor's wiki-synthesis selector filters the same list as the
chat model selector, and `wiki_ingest_batch.go` silently falls back to
`kb.SummaryModelID` when `wiki_config.synthesis_model_id` is unset. As a
result, operators have no way to declare a model as **specifically intended
for wiki synthesis** — the UI cannot distinguish "models suitable for wiki
page generation" from "models suitable for user-facing Q&A".

This matters in deployments that serve both general Q&A and wiki ingestion,
since the two roles usually want different models:

- **Chat (KnowledgeQA)**: latency-sensitive, smaller context, conversational.
- **Wiki synthesis**: long-context, batch-style, can afford a slower or
  cheaper model that's better at consolidating multiple chunks into a
  coherent wiki page.

Today the only way to surface "this model is meant for wiki synthesis" is
to name it something like `wiki-summary-gpt` and rely on naming
conventions, which is brittle.

This PR introduces a new `ModelTypeWikiSynthesis` enum value, parallel to
the existing `KnowledgeQA`, `Embedding`, `Rerank`, `VLLM`, and `ASR` types,
and wires it through Settings → Model Management and the KB editor so the
operator's intent becomes explicit.

The change is **additive and backwards-compatible**: existing KBs that
reference a `KnowledgeQA` model for wiki synthesis keep working, and the
ingest-time fallback to `kb.SummaryModelID` is left untouched. Deployments
that don't register any `WikiSynthesis`-typed model see no functional
change.

## What's in this PR

Three logical commits:

1. **`feat(types): introduce WikiSynthesis as a first-class model type`**
   Type system + backend handler wiring.
   - Add `ModelTypeWikiSynthesis` enum value in `internal/types/model.go`.
   - Declare it in `ModelTypes` for every chat-completion-capable provider
     (24 files). `GetDefaultURL`'s existing fallback to the chat URL means
     no `DefaultURLs` changes are needed.
   - Extend the frontend/backend type-name bridge in
     `internal/handler/model.go` so the UI can round-trip
     `"wikisynthesis"` ↔ `ModelTypeWikiSynthesis`.
   - Extend `findExistingModelID` (returns
     `kb.WikiConfig.SynthesisModelID`) and `buildConfigResponse` (emits a
     `wikiSynthesis` map entry) in `internal/handler/initialization.go`.
   - Add `'WikiSynthesis'` to the `ModelConfig.type` union in
     `frontend/src/api/model/index.ts` so the frontend TypeScript
     contract matches the backend enum.

2. **`feat(model-settings): surface WikiSynthesis as a first-class type in
   the model management UI`**
   - Add a dedicated tab and "Add Model" dropdown entry in
     `ModelSettings.vue`, plus a teal-ish `model-card__badge` style and
     `typeIcon` mapping.
   - Extend `ModelEditorDialog.vue`'s `Props.modelType` union.
   - Add `model.editor.description.wikisynthesis`,
     `modelSettings.typeShort.wikisynthesis`, and a per-type section
     across `zh-CN`, `en-US`, `ko-KR`, and `ru-RU` locales.

3. **`feat(kb-editor): route the wiki synthesis selector to the
   WikiSynthesis pool with KnowledgeQA fallback`**
   - Widen `ModelSelector.vue`'s `modelType` prop to accept an array, so
     the wiki synthesis selector can pass
     `['WikiSynthesis', 'KnowledgeQA']`. Single-string usages elsewhere
     (LLM / Embedding / Rerank / VLLM / ASR) are unaffected.
   - Render a `Wiki-only` outlined tag next to options whose
     `model.type === 'WikiSynthesis'`, so users can tell a purpose-
     registered wiki model apart from a chat model that also appears in
     the dropdown.
   - `KBModelConfig.vue` passes the new array prop and changes the
     "Add Model" action target from `knowledgeqa` to `wikisynthesis`.
   - Update `synthesisModelTip` i18n to make the fallback behavior
     explicit: "Falls back to the LLM model if not set" (the old
     "summary model" wording didn't match the actual KB field name).

The wiki ingest fallback in
`internal/application/service/wiki_ingest_batch.go` is deliberately left
intact — KBs that don't explicitly pick a wiki synthesis model continue
to fall back to the main LLM.

## Backwards compatibility

- **Existing KBs** with `wiki_config.synthesis_model_id` pointing at a
  KnowledgeQA-typed model continue to work; model lookup is by ID and
  ignores the type tag.
- **Existing deployments** without any WikiSynthesis-typed model
  registered see the wiki synthesis selector behave exactly as before —
  it shows the KnowledgeQA pool.
- **No schema migration** required. `ModelType` is a string enum; adding
  a value doesn't change any DB schema.
- **Wiki ingest fallback** to `kb.SummaryModelID` is preserved, so KBs
  that never pick a synthesis model still complete ingestion.

## Reproduce the new behavior

1. Restart the app after this PR is applied.
2. Go to **Settings → Model Management**. A new **Wiki Synth** tab
   appears between **Chat** and **Embedding**. The "Add Model" dropdown
   gets a corresponding entry.
3. Register a model under the **Wiki Synth** type (the same form as
   Chat; backed by the same chat-completion API).
4. Create a new knowledge base with Wiki enabled. Open **Models** in the
   editor: the **Wiki Synthesis Model** selector now shows the
   WikiSynthesis-typed model with a green **Wiki-only** tag, alongside
   any KnowledgeQA models as fallback options.

## Test plan

- [x] Backend: `go build ./internal/... ./cmd/...` — passes
- [x] Frontend: `npx vue-tsc --noEmit` — passes
- [x] Provider test suite (`go test ./internal/models/provider/...`) — passes
- [x] Register a WikiSynthesis model via the API (`POST /api/v1/models`
      with `"type":"WikiSynthesis"`) — succeeds.
- [x] Provider listing (`GET /api/v1/models/providers?model_type=wikisynthesis`)
      returns the same set of providers as `?model_type=chat`.
- [x] Settings UI shows the new tab, the new model card, and the
      type-coloured badge.
- [x] KB editor wiki synthesis dropdown shows both WikiSynthesis (with
      "Wiki-only" tag) and KnowledgeQA models.
- [x] Selecting either type and saving persists `wiki_config.synthesis_model_id`
      correctly.

## Non-goals / out of scope

- No change to the wiki ingest path; the fallback to `kb.SummaryModelID`
  is intentional to preserve current behavior.
- No automatic "default to first WikiSynthesis model" in the KB editor;
  operators pick explicitly. (The current behavior shows nothing
  pre-selected, matching the existing optional-field UX.)
- No model registry / catalog feature; this PR is strictly the type-
  system plumbing and UI wiring.

### DIFF
--- a/frontend/src/api/model/index.ts
+++ b/frontend/src/api/model/index.ts
@@ -8,7 +8,7 @@ export interface ModelConfig {
   id?: string;
   tenant_id?: number;
   name: string;
-  type: 'KnowledgeQA' | 'Embedding' | 'Rerank' | 'VLLM' | 'ASR';
+  type: 'KnowledgeQA' | 'WikiSynthesis' | 'Embedding' | 'Rerank' | 'VLLM' | 'ASR';
   source: 'local' | 'remote';
   description?: string;
   parameters: {
--- a/frontend/src/components/ModelEditorDialog.vue
+++ b/frontend/src/components/ModelEditorDialog.vue
@@ -280,7 +280,7 @@ interface ModelFormData {
 
 interface Props {
   visible: boolean
-  modelType: 'chat' | 'embedding' | 'rerank' | 'vllm' | 'asr'
+  modelType: 'chat' | 'wikisynthesis' | 'embedding' | 'rerank' | 'vllm' | 'asr'
   modelData?: ModelFormData | null
 }
 
@@ -960,7 +960,9 @@ const checkRemoteAPI = async () => {
 
     switch (props.modelType) {
       case 'chat':
-        // 对话模型（KnowledgeQA）
+      case 'wikisynthesis':
+        // 对话模型 (KnowledgeQA) 和 Wiki 合成模型 (WikiSynthesis) 都走 chat completion
+        // 接口，共用同一个连通性测试
         result = await checkRemoteModel({
           modelName: formData.value.modelName,
           baseUrl: formData.value.baseUrl,
--- a/frontend/src/components/ModelSelector.vue
+++ b/frontend/src/components/ModelSelector.vue
@@ -21,6 +21,7 @@
           <span class="model-name">{{ model.name }}</span>
           <t-tag v-if="model.is_builtin" size="small" theme="primary">{{ $t('model.builtinTag') }}</t-tag>
           <t-tag v-if="model.is_default" size="small" theme="success">{{ $t('model.defaultTag') }}</t-tag>
+          <t-tag v-if="model.type === 'WikiSynthesis'" size="small" theme="success" variant="outline">{{ $t('model.wikiSynthesisTag') }}</t-tag>
         </div>
       </t-option>
       
@@ -45,8 +46,14 @@ import { listModels, type ModelConfig } from '@/api/model'
 import { MessagePlugin } from 'tdesign-vue-next'
 import { useI18n } from 'vue-i18n'
 
+type SingleModelType = 'KnowledgeQA' | 'WikiSynthesis' | 'Embedding' | 'Rerank' | 'VLLM' | 'ASR'
+
 interface Props {
-  modelType: 'KnowledgeQA' | 'Embedding' | 'Rerank' | 'VLLM' | 'ASR'
+  // Single type or array of types. For example, the wiki synthesis selector
+  // passes ['WikiSynthesis', 'KnowledgeQA'] so WikiSynthesis-typed models
+  // surface first, and KnowledgeQA models act as a fallback pool for
+  // deployments that haven't registered any WikiSynthesis model.
+  modelType: SingleModelType | SingleModelType[]
   selectedModelId?: string
   disabled?: boolean
   placeholder?: string
@@ -72,10 +79,15 @@ const placeholderText = computed(() => {
   return props.placeholder || t('model.selectModelPlaceholder')
 })
 
+// Normalise the prop to an array so filtering can use includes() uniformly.
+const acceptedTypes = computed(() => {
+  return Array.isArray(props.modelType) ? props.modelType : [props.modelType]
+})
+
 // 监听 allModels 变化，自动过滤当前类型的模型
 watch(() => props.allModels, (newModels) => {
   if (newModels && Array.isArray(newModels)) {
-    models.value = newModels.filter(m => m.type === props.modelType)
+    models.value = newModels.filter(m => acceptedTypes.value.includes(m.type as SingleModelType))
   }
 }, { immediate: true })
 
@@ -96,7 +108,7 @@ const loadModels = async () => {
     const result = await listModels()
     // 前端按类型筛选模型
     if (result && Array.isArray(result)) {
-      models.value = result.filter(m => m.type === props.modelType)
+      models.value = result.filter(m => acceptedTypes.value.includes(m.type as SingleModelType))
     } else {
       models.value = []
     }
--- a/frontend/src/i18n/locales/en-US.ts
+++ b/frontend/src/i18n/locales/en-US.ts
@@ -2846,6 +2846,7 @@ export default {
       sourceRemote: 'Remote API',
       description: {
         chat: 'Configure large language models for conversations',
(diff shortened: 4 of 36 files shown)


lyingbug:
I'm not sure that introducing additional model types is a good way to extend this. Instead, it may be better to add tags or aliases to models, using soft constraints rather than hard restrictions. This approach could also provide better extensibility in the long run.

jackson-jia-914:
> I'm not sure that introducing additional model types is a good way to extend this. Instead, it may be better to add tags or aliases to models, using soft constraints rather than hard restrictions. This approach could also provide better extensibility in the long run.

Let me explain the issues users encounter in practical use:
1. Most users cannot distinguish between dialogue models and wiki synthesis models, so they mostly stick to the default settings or system recommendations.
2. When a dialogue model (usually a high-capability model, and sometimes a reasoning model) is selected, wiki content generation becomes extremely slow and consumes massive LLM resources.
After discussing with my team, we believe building a dedicated wiki model would be the optimal solution. Naturally, if users or system administrators do not configure it manually, the system will fall back to the dialogue model by default, which follows the existing code logic.
if you think this solution involves too many code changes, I can first add a wiki tag to mark the applicable scenarios for each model as you suggested. I’d like to hear your thoughts.

jackson-jia-914:
> I'm not sure that introducing additional model types is a good way to extend this. Instead, it may be better to add tags or aliases to models, using soft constraints rather than hard restrictions. This approach could also provide better extensibility in the long run.

Thanks for the suggestion. I'd like to pivot to the tag-based direction. Proposed shape:

1. Add a purposes []string field on Model (stored as a JSON column, defaults to empty array).
2. The model editor exposes a multi-select Purposes picker for KnowledgeQA-typed models. Initial options: qa, wiki-synthesis. Open-ended for future roles (e.g. summary, entity extraction).
3. The KB editor's wiki synthesis model selector still filters on KnowledgeQA type, but sorts/highlights models whose purposes contains wiki-synthesis to the top — a soft preference, not a hard filter.
4. No new ModelType, no changes to provider files, and the schema change is just one nullable JSON column — fully backward-compatible with existing models.

Does this direction work for you? Once you confirm, I'll rewrite the branch and force-push to replace the existing 3 commits.

lyingbug:
> > I'm not sure that introducing additional model types is a good way to extend this. Instead, it may be better to add tags or aliases to models, using soft constraints rather than hard restrictions. This approach could also provide better extensibility in the long run.
> 
> Thanks for the suggestion. I'd like to pivot to the tag-based direction. Proposed shape:
> 
> 1. Add a purposes []string field on Model (stored as a JSON column, defaults to empty array).
> 2. The model editor exposes a multi-select Purposes picker for KnowledgeQA-typed models. Initial options: qa, wiki-synthesis. Open-ended for future roles (e.g. summary, entity extraction).
> 3. The KB editor's wiki synthesis model selector still filters on KnowledgeQA type, but sorts/highlights models whose purposes contains wiki-synthesis to the top — a soft preference, not a hard filter.
> 4. No new ModelType, no changes to provider files, and the schema change is just one nullable JSON column — fully backward-compatible with existing models.
> 
> Does this direction work for you? Once you confirm, I'll rewrite the branch and force-push to replace the existing 3 commits.

This direction sounds great. I'm looking forward to your implementation.

jackson-jia-914:
Pivoted to the tag-based approach per your suggestion. New PR with the redesign: #1530.

In short:
- No new `ModelType`, no provider-file changes
- Adds `purposes []string` as a soft tag on `Model` (stored as a nullable JSONB column)
- Model editor exposes a multi-select for KnowledgeQA-typed models (initial options: `qa`, `wiki-synthesis`, open-ended for future roles)
- KB editor's wiki synthesis dropdown sorts purposes-tagged models to the top with a "Recommended for ..." badge, but untagged chat models remain selectable in the same dropdown — soft preference, not a hard filter

Will keep this PR open until #1530 review concludes, then close this one to keep history.

PR #1511, opened 2026-05-28
fix(deps): vendor-patch asynq v0.26.0 to skip undecodable task messages

toy0116:
## Summary

Adds a temporary `replace` directive in `go.mod` pointing `github.com/hibiken/asynq` at a locally-vendored copy under `internal/vendor/asynq`. The vendored copy has a one-line behavioral fix in `internal/rdb/rdb.go` that prevents the asynq recoverer goroutine from wedging when a single task message in Redis cannot be decoded.

The exact same fix is proposed upstream in **[hibiken/asynq#1140](https://github.com/hibiken/asynq/pull/1140)**. This PR is intended to be temporary — once asynq#1140 lands in a release and we bump asynq above v0.26.0, the `replace` directive and the entire `internal/vendor/asynq` tree should be removed.

## Bug repro

Observed on a production WeKnora deployment running asynq v0.26.0:

```
asynq: ... WARN: recoverer: could not list lease expired tasks:
  INTERNAL_ERROR: cannot decode message: proto: cannot parse invalid wire-format data
```

The recoverer runs on a 1-minute timer and aborts on the first undecodable task message it sees in the lease set, so no lease-expired task ever gets retried or archived. New tasks can still be enqueued and processed, but anything that lost its lease (worker restart, lease extension miss, etc.) is stuck forever.

In our case, this left **~50 newly uploaded documents permanently stuck in `parse_status='pending'`** because the parser tasks were waiting for a worker dispatch that the recoverer was supposed to unblock. The recoverer had been silently logging this error every minute for ~3 days before we caught it. Manual `DEL` of every `asynq:{...}:archived` Redis key was required to unwedge.

## Fix

`RDB.ListLeaseExpired` (and the analogous loop in `RDB.ReadAggregationSet`) now logs a warning and `continue`s past undecodable messages instead of returning an error that abort the whole batch:

```go
msg, err := base.DecodeMessage([]byte(s))
if err != nil {
    log.Printf("asynq[rdb.X]: skipping undecodable message ...")
    continue
}
msgs = append(msgs, msg)
```

The hot-path single-task `Dequeue` at `rdb.go:381` is intentionally left as-is: that one should propagate errors so the caller can retry.

## Why vendor + `replace` rather than wait for upstream

asynq's open PR list has 50+ entries and the project's release cadence has been slow (latest tag `v0.26.0` is from Feb 2026). Vendoring under `internal/vendor/` is a pragmatic short-term path so WeKnora users on `v0.5.x` are not exposed to a footgun that can wedge their worker indefinitely. The vendored tree is a verbatim copy of asynq v0.26.0 with only two changed sites in `internal/rdb/rdb.go` — diff against upstream is small and easy to audit.

## Test

- `go build ./...` clean
- WeKnora worker (rebuilt from this branch) ran for several hours under real traffic with a known-corrupt task message planted in Redis: the offending message is skipped with one warning log per recoverer tick, the rest of the queue makes progress as expected, and no document gets stuck on the parse path.
- Upstream PR (hibiken/asynq#1140) carries a unit test `TestListLeaseExpiredSkipsUndecodable` that locks the behavior of the patch.

## Cleanup checklist (for when asynq#1140 lands)

1. Bump `github.com/hibiken/asynq` to the release containing the fix.
2. Remove the `replace` directive at the bottom of `go.mod`.
3. Delete `internal/vendor/asynq/`.
4. `go mod tidy`.


lyingbug:
Can you tell what content is causing the decode failure? Maybe we can avoid this kind of issue by controlling what gets written into the queue.  @toy0116

toy0116:
> Can you tell what content is causing the decode failure? Maybe we can avoid this kind of issue by controlling what gets written into the queue. @toy0116

Its 10k size MD doc

lyingbug:
So you were maintaining the Markdown documents using the manual upload (editable) approach, rather than uploading files directly?

toy0116:
I'm uploading md doc directly, that time I uploaded 7 files, each around 10-20kb.

PR #1528, opened 2026-05-28
fix: parse xlsx files with malformed styles

wolfkill:
## Summary
- add a raw `.xlsx` XML fallback when pandas/openpyxl rejects workbook styles
- preserve the existing row-to-chunk output format by treating the first row as headers
- add a regression test for the openpyxl `expected Fill` failure path

This addresses the Excel parsing failure reported in #612. The MCP timeout mentioned there appears separate.

## Testing
- `uv run --project docreader python -m unittest discover -s docreader/tests -p 'test_excel_parser.py'`
- `uv run --project docreader python -m unittest discover -s docreader/tests`
- `git diff --check`

### DIFF
--- a/docreader/parser/excel_parser.py
+++ b/docreader/parser/excel_parser.py
@@ -6,8 +6,12 @@ structured Document objects with text content and chunks. It supports multiple
 sheets and handles various Excel formats using pandas.
 """
 import logging
+import posixpath
+import re
+import xml.etree.ElementTree as ET
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


lyingbug:
Thanks! This has been open since May and now conflicts with main, and #612 is already closed. Hand-rolling an xlsx XML parser as the fallback is also a lot to maintain. Closing. If you still hit a malformed-styles workbook on current main, please open an issue with a sample file.

PR #1530, opened 2026-05-28
feat(models): add purpose tags as a soft recommendation for model role

jackson-jia-914:
## Summary

Adds a soft-tag mechanism for declaring what role a model is *recommended for*, complementing the existing `ModelType` axis. This is the tag-based redesign of #1509, pivoted per @lyingbug's feedback on that PR: rather than introducing a new `ModelType` enum value (hard constraint), a model now carries an open-ended `purposes []string` that selectors interpret as a sort/highlight hint while still allowing untagged models in the same dropdown (soft constraint, better extensibility).

## Background

Operators today cannot distinguish "chat models that are good for user-facing Q&A" from "chat models that are good for wiki page generation" — both speak the same `KnowledgeQA` chat-completion API contract. The practical pain we hit in production:

1. Users can't tell the two kinds of chat models apart, so they default to whichever model is configured as the system LLM.
2. When the system LLM happens to be a high-end reasoning model, wiki ingestion (long-context Map-Reduce) becomes extremely slow and burns disproportionate token budget.

`ModelType` already encodes the *API contract* a model speaks (`Embedding` / `Rerank` / `KnowledgeQA` / `VLLM` / `ASR`). What was missing was a way to declare the *intended role within a contract*. `purposes` fills exactly that gap, and is intentionally open-ended (string slice rather than enum) so future roles — summary, entity extraction, agent reasoning, image captioning, … — can land without schema or enum churn.

## What's in this PR

Three commits:

1. **`feat(types): add Purposes soft-tag field on Model`**
    - `ModelPurposes []string` with `Scan`/`Value` JSON column impl
    - `Model.Purposes` GORM field
    - `CreateModelRequest.Purposes` and `UpdateModelRequest.Purposes` (the update field is a pointer so omitting it leaves stored tags untouched, sending `[]` clears them, sending `[...]` replaces)
    - `ModelResponse.Purposes` with `omitempty` so untagged models keep the field absent in API responses
    - Migration `000057_models_purposes` — nullable JSONB column, `IF NOT EXISTS` guarded, idempotent and reversible
    - Matching TS type alignment in `frontend/src/api/model/index.ts`

2. **`feat(model-settings): expose Purposes multi-select for chat models`**
    - In the model editor dialog, `KnowledgeQA`-typed (chat) models get a new "Purposes" multi-select with initial options `qa` and `wiki-synthesis`
    - The picker is **hidden** for `Embedding` / `Rerank` / `VLLM` / `ASR` since those are already differentiated by API contract
    - `ModelSettings` forwards the selected purposes through `createModel` / `updateModel` and back-fills on edit
    - For non-chat types, sends `[]` explicitly so a type-switch deterministically clears stale tags
    - i18n keys added across all four locales (zh-CN, en-US, ko-KR, ru-RU); `synthesisModelTip` wording also updated to describe the new behavior

3. **`feat(kb-editor): soft-prefer purposes-tagged models in the wiki synthesis selector`**
    - `ModelSelector` gains an optional `preferredPurpose` prop
    - When set, models whose `purposes` contain that value sort to the top and render a green outlined "Recommended for ..." tag
    - Untagged models of the same type stay in the dropdown and remain selectable — soft preference, **not** a hard filter
    - `KBModelConfig` wires `preferred-purpose="wiki-synthesis"` on the wiki synthesis dropdown
    - No backend or fallback behavior changes: KBs without a synthesis model continue to fall back to the chat model at ingest time

## Backwards compatibility

- **Existing models** have `purposes = NULL`; the column is nullable and `omitempty` in responses, so untagged models behave exactly as before (no usage hint, selectors fall back to type-based ordering).
- **Existing KBs** that reference a `KnowledgeQA` model for wiki synthesis keep working — lookup is by model ID and ignores tags.
- **No new `ModelType`**, no provider-file changes, no API breaking changes. The DTO addition is additive.
- **Migration is one column add**, guarded with `IF NOT EXISTS`, cleanly rollback-able.

## Visual

A chat model registered with both purposes ticked:

`Model editor → Chat → Purposes ☑ User Q&A  ☑ Wiki synthesis`

When opening the KB editor → Wiki synthesis dropdown:

```
✓ Qwen3-27B-wiki         [Recommended for Wiki synthesis]   ← sorted top
✓ kimi-k2          [Built-in] [Default]
+ Add a model in Settings
```

The tagged model floats to the top with a green outlined badge; the untagged `kimi-k2` is still selectable in the same dropdown.

## Test plan

- [x] `go build ./internal/... ./cmd/...` — passes
- [x] `npx vue-tsc --noEmit` — passes
- [x] `go test ./internal/handler/... ./internal/handler/dto/...` — passes
- [x] Manual: register a chat model with `purposes=["wiki-synthesis"]` via the UI → persists, reopen editor reflects the checked boxes
- [x] Manual: KB editor wiki synthesis dropdown places the tagged model on top with the "Recommended for Wiki synthesis" tag; untagged chat models remain selectable
- [x] Manual: editor for `Embedding` / `Rerank` / `VLLM` / `ASR` does **not** show the Purposes picker
- [x] DB roundtrip: `purposes` written via `UpdateModel` is retrievable via `GET /api/v1/models`

## Non-goals / out of scope

- No automatic "default to first wiki-synthesis-tagged model" in the KB editor — operators still pick explicitly, matching the existing optional-field UX.
- No backend-side filtering on `purposes` at ingest time — purposes are advisory metadata for selectors, not a runtime gate.
- The wiki ingest fallback to the chat model (when `wiki_config.synthesis_model_id` is unset) is intentionally preserved.

## Relationship to #1509

This PR supersedes the approach in #1509. Closing #1509 in favor of this one once review here lands. Discussion thread on [#1509](https://github.com/Tencent/WeKnora/pull/1509) has the original context.

### DIFF
--- a/frontend/src/api/model/index.ts
+++ b/frontend/src/api/model/index.ts
@@ -37,6 +37,10 @@ export interface ModelConfig {
   is_default?: boolean;
   is_builtin?: boolean;
   status?: string;
+  // Soft tag list of intended usage roles for the model. Common values:
+  // "qa" (user-facing chat) and "wiki-synthesis" (wiki page generation).
+  // Open-ended — new roles can be added without API changes.
+  purposes?: string[];
   // Per-field configured? metadata from the main response. Absent for
   // builtin models.
   credentials?: Record<ModelCredentialField, { configured: boolean }>;
--- a/frontend/src/components/ModelEditorDialog.vue
+++ b/frontend/src/components/ModelEditorDialog.vue
@@ -240,6 +240,16 @@
         </div>
       </div>
 
+      <!-- Chat: purpose tags (soft recommendations consumed by KB editor) -->
+      <div v-if="modelType === 'chat'" class="form-item">
+        <label class="form-label">{{ $t('model.editor.purposesLabel') }}</label>
+        <t-checkbox-group v-model="formData.purposes">
+          <t-checkbox value="qa">{{ $t('model.editor.purposes.qa') }}</t-checkbox>
+          <t-checkbox value="wiki-synthesis">{{ $t('model.editor.purposes.wikiSynthesis') }}</t-checkbox>
+        </t-checkbox-group>
+        <p class="form-desc">{{ $t('model.editor.purposesDesc') }}</p>
+      </div>
+
     </t-form>
   </SettingDrawer>
 </template>
@@ -283,6 +293,8 @@ interface ModelFormData {
   supportsVision?: boolean
   // 自定义 HTTP 请求头（类似 OpenAI Python SDK 的 extra_headers）
   customHeaders?: CustomHeaderItem[]
+  // 用途软标签，目前仅 chat 类型暴露给用户：qa / wiki-synthesis
+  purposes?: string[]
 }
 
 interface Props {
@@ -570,7 +582,8 @@ const formData = ref<ModelFormData>({
   interfaceType: 'ollama',
   isDefault: false,
   supportsVision: false,
-  customHeaders: []
+  customHeaders: [],
+  purposes: []
 })
 
 const rules = computed(() => ({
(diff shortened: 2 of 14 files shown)


jackson-jia-914:
The style is as follows：
image
image

Issue #1513, opened 2026-05-28
[Feature]: 希望可以支持腾讯云知识引擎原子能力的rerank配置


phihere:
### Affected Component

Frontend UI

### Problem Description

https://cloud.tencent.com/document/product/1772/115339
知识引擎原子力的有rerank 接口

背景是，我们支持的好多项目都想在腾讯云解决所有配置，唯独rerank需要单独走其他厂商购买，更希望可以一站式解决所有模型配置能力

### Proposed Solution

希望可以一站式解决所有模型配置能力，比如支持腾讯云的rerank模型配置

### Impact

Many users would benefit

### Confirmation

- [x] I have searched existing issues and confirmed this is a new request
- [x] I understand this request may need discussion and evaluation

PR #1533, opened 2026-05-28
fix: support lkeap rerank provider

wolfkill:
## Summary
- add a Tencent Cloud LKEAP RunRerank implementation with TC3 request signing
- register LKEAP as a rerank-capable provider with the official Cloud API endpoint
- carry AppSecret through rerank test-connection requests and model config fallback

## Motivation
Addresses #1513. The existing LKEAP provider only covered OpenAI-compatible KnowledgeQA models, so rerank models fell back to the generic OpenAI-style /rerank path instead of Tencent Cloud RunRerank.

## Validation
- `CGO_LDFLAGS='-lc++' go test ./internal/models/rerank -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/models/provider -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/handler -count=1`
- `git diff --check`

### DIFF
--- a/internal/handler/initialization.go
+++ b/internal/handler/initialization.go
@@ -1532,13 +1532,14 @@ type ModelTestRequest struct {
 	ModelName     string            `json:"modelName" binding:"required"`
 	BaseURL       string            `json:"baseUrl"`
 	APIKey        string            `json:"apiKey"`
+	AppSecret     string            `json:"appSecret,omitempty"`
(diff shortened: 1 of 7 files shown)


wolfkill:
Closing this draft to avoid carrying an unreviewed provider-specific rerank implementation without live LKEAP API verification. This can be reopened or rebuilt later with real endpoint evidence.

Issue #1401, opened 2026-05-20
[Feature]: Support asymmetric embedding via task_type / input_type parameters


pty819:
### Affected Component

Backend Service & API

### Problem Description

    Background

    Many modern embedding models (especially Jina v5 retrieval models) are asymmetric — they produce different embeddings for queries and documents depending on the task or input_type parameter. This is critical for retrieval scenarios because queries and documents come from different semantic distributions.

    For example, Jina's jina-embeddings-v5-text-small-retrieval uses:
    - task: "retrieval.query" → prepends "Query: " before encoding
    - task: "retrieval.passage" → prepends "Document: " before encoding

    Using the wrong task type can significantly degrade retrieval performance.

    Current WeKnora Implementation

    After reviewing the source code, WeKnora currently does not support asymmetric embeddings:

    Embedding Provider Request Structures

    OpenAI-compatible (internal/models/embedding/openai.go):
    go
    type OpenAIEmbedRequest struct {
        Model              string json:"model"
        Input              []string json:"input"
        EncodingFormat     string json:"encoding_format,omitempty"
        TruncatePromptTokens int json:"truncate_prompt_tokens,omitempty"
    }

    Jina (internal/models/embedding/jina.go):
    go
    type JinaEmbedRequest struct {
        Model     string json:"model"
        Input     []string json:"input"
        Truncate  bool json:"truncate,omitempty"
        Dimensions int json:"dimensions,omitempty"
    }

    Neither structure includes task, task_type, or input_type fields. A full codebase search for task_type, query_prefix, document_prefix, or prefix returns zero matches.

    Consequences

    1. Jina retrieval models lose asymmetric optimization — both queries and documents are embedded without the task parameter, effectively treating them as symmetric. This can reduce recall by 5-15% on retrieval benchmarks.
    2. No query/document distinction — the Embed() interface is identical for both query vectors (during search) and document vectors (during indexing).
    3. Cannot leverage Jina gateway's automatic prefix feature — even if a Jina gateway is configured to auto-prepend prefixes based on task_type, WeKnora cannot send that parameter.

    OpenViking Implementation (Reference)

    OpenViking (ByteDance's context database) handles asymmetric embeddings properly. Here's how:

    Architecture

    Search query  → embed_compat(embedder, query, is_query=True)
    Document text → embed_compat(embedder, doc, is_query=False)

    The is_query boolean flows through the entire call chain, from search/retrieval down to the embedding provider.

    Jina Provider

    python
[...]

PR #1539, opened 2026-05-29
fix: support asymmetric embedding parameters

wolfkill:
## Summary
- add optional embedding request roles so query embeddings and document embeddings can carry different provider parameters
- support OpenAI-compatible `input_type` via `extra_config.query_input_type` / `extra_config.document_input_type`
- support Jina `task` and `task_type` via embedding model `extra_config`, and document the supported keys

Addresses #1401.

## Validation
- `CGO_LDFLAGS='-lc++' go test ./internal/models/embedding`
- `CGO_LDFLAGS='-lc++' go test ./internal/application/service/retriever ./internal/application/service`
- `git diff --check`

Note: I validated outgoing request payloads with local HTTP transport tests. I did not call a live Jina or OpenAI-compatible embedding provider.


wolfkill:
Closing this draft because the asymmetric embedding parameter change touches many embedding providers and was validated with local payload tests only, not live Jina/OpenAI-compatible providers. Better to avoid leaving broad unverified provider behavior open.

Issue #1440, opened 2026-05-22
Phase 3: OpenSearch k-NN Native Vector Search Driver


ochanism:
## Context

This is Phase 3 of the multi-store vector DB architecture proposal (#913), following Phase 1 (#921) which introduced VectorStore as a first-class entity and Phase 2 (#993) which bound KBs to specific VectorStores with multi-store fan-out search and engine-aware score normalization.

Phase 3 adds a native OpenSearch k-NN retriever driver to the multi-store ecosystem. The existing Elasticsearch driver uses `script_score` with `cosineSimilarity` — a brute-force O(N) scan that degrades to seconds at 100K+ documents. OpenSearch's k-NN plugin provides native HNSW-indexed approximate-nearest-neighbor search (O(log N)), offering 10×–500× latency improvements at scale while reusing the rest of WeKnora's storage, chunking, and fusion pipeline unchanged.

## Goals

1. **Native k-NN driver via opensearch-go v4** — `internal/application/repository/retriever/opensearch/` package implementing `RetrieveEngineRepository`, using the cluster's k-NN plugin with Lucene (default) or Faiss HNSW engines
2. **OpenSearch 2.x + 3.x cross-version support, single SDK** — the k-NN wire format is identical across both major versions; a single `opensearch-project/opensearch-go v4.x` dependency suffices. Hard floor: OS 2.4 (Lucene HNSW GA). Primary tested target: OS 3.3.2
3. **Per-dimension index naming with lazy initialization** — `<base>_<dim>` alias-backed indexes, lazy-initialized on first use via `sync.Once`, matching the Qdrant collection pattern. Different embedding-model KBs bound to the same store get isolated indexes safely; no PUT happens at container init
4. **Engine-aware score normalization extension** — new `EngineAwareNormalizer` case for OpenSearch's cosine-shifted `[0, 2]` raw score → `[0, 1]`. Multi-store fan-out from Phase 2 treats `opensearch` as any other engine
5. **Index alias from day one** — actual index is `<alias>_v1`, alias points at it. Enables mapping evolution and rolling reindex without downtime
6. **TLS hardening** — pinned `MinVersion: TLS12, InsecureSkipVerify: false` by default; new `ConnectionConfig.InsecureSkipVerify` field (backward-compat default `false`) for self-signed dev clusters
[...]

PR #1542, opened 2026-05-29
feat(retriever): activate OpenSearch k-NN driver (PR 3 of 3)

ochanism:
## Summary

Flip the gate for the OpenSearch k-NN driver scaffolded across #1445 / #1481 / #1482 by way of [#1440](https://github.com/Tencent/WeKnora/issues/1440). PR 1 (type prep) + PR 2a (driver skeleton) + PR 2b (read/write paths) all landed as **gated dead code**; this PR wires every activation surface so `opensearch` becomes a registerable VectorStore engine, replaces the remaining `ErrFeatureNotEnabled` stubs with real sync implementations, adds the driver-owned audit, and ships the schema-driven frontend renderer.

**This is the activation PR**: after merge, registering an `engine_type=opensearch` VectorStore (DB store or `RETRIEVE_DRIVER=opensearch` env path) creates a live, fan-out-capable k-NN engine. Before this PR the same attempt fails with `"not a valid engine type"`.

Async operations (task-polled `_reindex` above the result window, the lifted `DeleteBy*` cap with bounded polling) and the engine-agnostic `vector_store.*` audit are **intentionally out of scope** and tracked as the next follow-up — the sync paths here cover typical KB sizes (pagination bounded by `max_result_window`, default 10000).

## Context

Part of [#1440](https://github.com/Tencent/WeKnora/issues/1440) — Phase 3 (OpenSearch k-NN Native Vector Search Driver) of the [multi-store roadmap](https://github.com/Tencent/WeKnora/issues/913).

Phase 3 PR sequence (this is PR 3 of 3):

| # | PR | Status |
|---|----|--------|
| 1 | type prep (gated, no activation) | MERGED ([#1445](https://github.com/Tencent/WeKnora/pull/1445)) |
| 2a | driver skeleton + interface stubs (gated dead code) | MERGED ([#1481](https://github.com/Tencent/WeKnora/pull/1481)) |
| 2b | driver read/write implementations (gated dead code) | MERGED ([#1482](https://github.com/Tencent/WeKnora/pull/1482)) |
| **3** | **this PR — activation switch + sync CopyIndices/BatchUpdate + driver audit + frontend renderer** | **here (ready for review)** |

This branch is rebased onto `upstream/main` HEAD; the diff against `main` collapses to just the activation payload below.

Depends on Phase 1 ([#921](https://github.com/Tencent/WeKnora/issues/921), MERGED), Phase 2 ([#993](https://github.com/Tencent/WeKnora/issues/993), all 5 PRs MERGED), and Phase 3 PRs 1/2a/2b (all MERGED).

## Changes at a glance

35 files, +1781/−169. The gate flip touches the type system, the DI container, the service layer, the driver package, and the frontend; the guard tests from PR 1 / PR 2 are replaced by their positive counterparts.

- **Activation wiring** — `internal/types`: `validEngineTypes` / `GetVectorStoreTypes` (with HNSW bounds, `knn_engine` enum, `Immutable` hints) / `retrieverEngineMapping` / `buildEnvStoreForDriver` now all recognise `"opensearch"`. `IndexConfig` grows four `omitempty` HNSW fields (`HNSWM` / `HNSWEFConstruction` / `HNSWEFSearch` / `KNNEngine`), keeping every other engine's serialised config byte-identical. `internal/container`: `createOpenSearchEngine` + the `createEngineServiceFromStore` switch case + the `RETRIEVE_DRIVER=opensearch` path in `initRetrieveEngineRegistry`; `NewEngineFactory` closes over the `AuditLogService` (the `EngineFactory` type itself is unchanged). `transport.go`: `NewOpenSearchClient` is exported so the factory and env path can build the TLS-hardened client.

- **Service-layer validation** — `validateOpenSearchIndexConfig` enforces the HNSW caps (`m` 2–100, `ef_construction` 2–4096, `ef_search` 1–10000, `knn_engine ∈ lucene|faiss`); shards/replicas stay with the flat `ValidateIndexConfig`. **Create-only**: `UpdateStore` mutates the name only, so existing KB bindings stay safe. `vectorstore_healthcheck.go` gains a `testOpenSearchConnection` case (version + k-NN plugin probe) so `CreateStore`'s connectivity check accepts opensearch instead of returning 400.

- **Sync implementations (`stubs.go` shrinks)** — `CopyIndices` (`copy.go`) mirrors the Elasticsearch / Qdrant pattern (search → `BatchSave` with the `source_id` remap for generated questions), so dim/keyword routing and the `source_id` contract come from `BatchSave` for free; native `_reindex` is deliberately not used. `BatchUpdateChunkEnabledStatus` / `BatchUpdateChunkTagID` (`bulk_update.go`) group input by target value and issue one `_update_by_query` per group over the cross-dim `<base>_*` pattern — caller values flow through **bound script params only**, never string-interpolated into the Painless source. `inspectByQueryResponse` (`byquery.go`) mirrors `inspectBulkResponse`: the full failure reason goes to the debug log only; the returned error carries the bounded id + type.

- **Driver-owned audit (DIP)** — a new `opensearch.AuditSink` interface (with `nopSink` + the `WithAuditSink` functional option) lets the driver emit `opensearch.index_created` / `opensearch.reindex_executed` without importing any service package. `NewRepository` takes opts, so existing 4-arg call sites keep compiling. `internal/container/audit_sink.go` bridges the sink to `AuditLogService`; when the context carries no tenant (env-path boot registration) the adapter **skips the emit with a warning** rather than writing `tenant_id=0`, which would collide with the system-scope sentinel.

- **Frontend + i18n** — `FieldSchema` (`frontend/src/api/vector-store.ts`) gains `min`/`max`/`enum`/`immutable`. `VectorStoreSettings.vue` is now schema-driven: a closed `enum` renders a `t-select`; number inputs use the schema's `:min`/`:max` and fall back to the legacy replica-vs-shard heuristic only when the schema does not pin them; a danger-coloured warning fires when `insecure_skip_verify` is toggled on (the switch and warning are wrapped in a vertical stack so the warning sits on its own row). i18n labels for the HNSW fields / `knn_engine` / `insecure_skip_verify` plus the warning copy in en-US, ko-KR, zh-CN, ru-RU.

- **Dev infra + docs** — `docker-compose.dev.yml` gains an `opensearch` profile (single-node 3.3.2, security plugin disabled for dev only). OpenSearch Dashboards lives in a **separate, opt-in `opensearch-ui` profile** so the heavy UI container is never forced up alongside the cluster (the driver e2e is fully curl-verifiable against `:9200`). `docs/dev/opensearch-integration-test.md` covers the end-to-end exercise and the single-node guidance (set `replicas=0` to keep the cluster Green).

- **Gating-guard tests flipped** — the "OpenSearch is NOT in `validEngineTypes` / mapping / types list / env builder / stubs" guard tests from PR 1 / PR 2 are replaced by their positive counterparts. The test suite was the activation checklist; the activation flip is its diff. New tests across `internal/types`, `opensearch`, `service`, and `container` — including a full end-to-end env-path test that exercises `initRetrieveEngineRegistry` with `RETRIEVE_DRIVER=opensearch` against an `httptest` cluster.

### Out of scope (next follow-up)

- **Async ops** — task-polled `_reindex` for copies above `max_result_window`, the lifted `DeleteBy*` cap with a hard ceiling + bounded context, and `_tasks` raw decode. The sync paths here are capped (delete 1000 IDs/call, copy bounded by the result window) and reject oversize batches with `ErrBatchTooLarge`.
- **`vector_store.*` audit** — engine-agnostic, cross-cutting; lands as a separate PR. This PR ships only the two driver-owned `opensearch.*` events.

## Backward compatibility

- Additive everywhere. `IndexConfig`'s new HNSW fields are `omitempty`, so other engines' serialised config stays byte-identical.
- Existing Elasticsearch / Qdrant / Milvus / Weaviate / Doris / TencentVectorDB stores are untouched.
- No migrations. The k-NN index is created lazily on first KB use, not on merge.
- The PR 1 normalizer case for OpenSearch becomes reachable for the first time here (cosine score passthrough in `[0, 1]`).

## Test plan

Verified on the rebased branch (single commit on `upstream/main@bb202a20`):

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `gofmt -l` — clean on all touched files
- [x] `go test ./internal/application/repository/retriever/opensearch/... ./internal/types/... ./internal/container/...` — passes (incl. the env-path activation e2e against an `httptest` cluster)
- [x] `go test ./...` — the only failures are `TestOssEnsureBucket_CreateFails` (Aliyun OSS endpoint reachability) and the `doris` SQL-shape tests, both pre-existing on `upstream/main` and untouched by this PR
- [x] **Manual local integration** (per `docs/dev/opensearch-integration-test.md`, single-node OpenSearch 3.3.2): registered a DB store, ingested documents, confirmed per-dim index + alias creation (`weknora_<store>_<dim>_v1` + alias), verified hybrid retrieval (vector k-NN + keyword BM25 fused via service-layer RRF) returns `[0, 1]`-normalised scores, and confirmed `_id == chunk_id` on every indexed document


lyingbug:
LGTM

PR #1556, opened 2026-05-30
feat(auth): make JWT token TTL configurable via environment variables

nixidexiangjiao:
## Description

JWT access/refresh token lifetimes were previously hardcoded (access: 24h, refresh: 7d) in `generateTokensForTenant`. This PR makes both configurable through environment variables, with full backward compatibility — when unset, the original defaults are used.

- Add `JWT_ACCESS_TOKEN_TTL` and `JWT_REFRESH_TOKEN_TTL` env vars (default `24h` / `7d`).
- Values accept Go duration syntax **plus a `d` (day) suffix** (e.g. `30m`, `24h`, `30d`), since the stdlib `time.ParseDuration` only supports units up to `h`. Invalid or non-positive values log a warning and fall back to the default.
- Compute the issuing timestamp once so the JWT `exp` claim and the persisted `AuthToken.ExpiresAt` stay in sync (previously they used separate `time.Now()` calls).
- Wire the new vars through `docker-compose.yml` and the Helm chart (`values.yaml` + `templates/app.yaml`, injected only when non-empty).

The refresh flow (`RefreshToken`) reuses `GenerateTokens`, so it picks up the configured TTLs automatically — no separate change needed.

## Type of Change

- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test
- [x] 🔧 Configuration / Build / CI

## Related Issue

Fixes #

## Testing

- Added `internal/application/service/token_ttl_test.go` with table-driven coverage:
  - `parseDuration`: `d` (day) unit (`30d`, `7d`, `0.5d`), native formats (`24h`, `30m`, `1h30m`), and error cases (`abc`, `d`, empty).
  - `parseDurationEnv`: empty/whitespace → fallback, valid overrides (`30d`, `48h`), and invalid/zero/negative → fallback.
- `go test ./internal/application/service/ -run 'TestParseDuration|TestParseDurationEnv' -v` — all pass.
- `go vet ./internal/application/service/` — clean.
- `helm template` verified: TTL env vars are injected when set, and omitted entirely when left blank (matching the docker-compose `${VAR:-}` behavior).

## Checklist

- [x] Self-reviewed the code

## Screenshots / Recordings

N/A — no user-visible UI changes.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -650,3 +650,7 @@ DOCREADER_TRANSPORT=grpc
 # Background sweep that recovers knowledge stuck in "processing" past
 # DocumentProcessTimeout + 10m. Set to "false" to opt out (not recommended).
 # WEKNORA_HOUSEKEEPING_ENABLED=true
+
+# JWT token validity period (in Go duration format, e.g., 24h / 168h); leave blank to use default values (access 24h / refresh 7d)
+JWT_ACCESS_TOKEN_TTL=
+JWT_REFRESH_TOKEN_TTL=
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,7 +173,7 @@ services:
       - SSRF_WHITELIST_EXTRA=${SSRF_WHITELIST_EXTRA:-searxng}
       - CONCURRENCY_POOL_SIZE=${CONCURRENCY_POOL_SIZE:-5}
       - JWT_SECRET=${JWT_SECRET:-}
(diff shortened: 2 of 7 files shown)


lyingbug:
doc-content page is not need modify in this pr

nixidexiangjiao:
> doc-content page is not need modify in this pr

doc-content page is reverted already

PR #1588, opened 2026-06-03
fix: skip credential validation when editing data source

langcaiye:
## Summary
- skip required credential field validation when editing a data source that already has credentials configured
- keep validation unchanged for creating a data source or explicitly replacing credentials

## Issue
- I did not find a matching open issue for this bug.

## Test
- npm run build
- git diff --check


langcaiye:
修复的是个最近引入的很明显的bug，点击【测试链接】，直接没有认证信息
可以优先处理一下吗 @lyingbug 
image

PR #1593, opened 2026-06-04
fix(dev): use writable local storage path

langcaiye:
## Summary
- make dev-app use a repo-local writable storage directory when LOCAL_STORAGE_BASE_DIR is unset or still set to the Docker default /data/files
- preserve explicitly configured local storage paths
- keep Docker Compose deployments unchanged

## Motivation
The default /data/files path is writable inside the Docker app container because a volume is mounted there, but it can be read-only or missing when the backend runs directly on the host via make dev-app. File ingestion then fails with errors like mkdir /data: read-only file system.

## Test
- bash -n scripts/dev.sh
- git diff --check

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -352,6 +352,16 @@ start_app() {
     export OTEL_EXPORTER_OTLP_ENDPOINT=localhost:4317
     export NEO4J_URI=bolt://localhost:7687
     export QDRANT_HOST=localhost
+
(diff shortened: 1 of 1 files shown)


langcaiye:
解决飞书同步时的这类问题：
  all fetched items failed during sync (80/80): XXXX: failed to create directory: mkdir /data: read-only file system

PR #1605, opened 2026-06-04
fix(schema): expand knowledge source length

langcaiye:
## Description

`knowledges.source` currently uses `VARCHAR(128)`, but file and URL knowledge sources can be longer than that in normal deployments. Object storage URLs are a common example: bucket domains, path prefixes, timestamps, and URL-encoded non-ASCII filenames can easily exceed 128 characters.

This PR expands the bounded column length to `VARCHAR(2048)` instead of using `TEXT`, keeping the field size explicit while allowing typical long URLs.

## Changes

- Updates `Knowledge.Source` GORM metadata to `varchar(2048)`.
- Updates initial schemas for versioned/Postgres, MySQL, SQLite, and ParadeDB migrations.
- Adds versioned migration `000058_expand_knowledge_source` for existing deployments.
- Adds a repository schema test that verifies the generated `knowledges.source` column type and persists a long object-storage-style URL.

## Testing

- `git diff --check`
- `go test ./internal/types ./internal/application/repository`

### DIFF
--- a/internal/application/repository/knowledge_finalize_test.go
+++ b/internal/application/repository/knowledge_finalize_test.go
@@ -27,7 +27,7 @@ CREATE TABLE IF NOT EXISTS knowledges (
     type VARCHAR(50) NOT NULL DEFAULT '',
     title VARCHAR(255) NOT NULL DEFAULT '',
     description TEXT,
-    source VARCHAR(128) NOT NULL DEFAULT '',
+    source VARCHAR(2048) NOT NULL DEFAULT '',
     parse_status VARCHAR(50) NOT NULL DEFAULT 'unprocessed',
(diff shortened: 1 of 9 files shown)


langcaiye:
我们在对象存储场景遇到了 `knowledges.source` 长度不够的问题。对象存储 URL 里包含 bucket 域名、路径前缀、时间戳，以及 URL encode 后的中文文件名，超过 128 字符很常见。

这里没有改成 TEXT，而是选择 VARCHAR(2048)，主要是希望保持字段长度有明确上限，避免给 MySQL/DBA 治理带来额外复杂度，同时 2048 对常见 URL/对象存储地址已经比较充足。

Issue #58, opened 2025-08-13
git pull今天最新的，在docker构建过程中报错了


dfelectric:
=> ERROR [docreader builder  3/13] RUN apt-get update && apt-get install -y     gcc     python3-dev     libjpeg-dev     zlib1g-dev     libpq-dev     libffi-dev     libgl1-mesa-glx     libglib2.0-0    9.9s
 => ERROR [docreader runner  3/15] RUN apt-get update && apt-get install -y     libjpeg62-turbo     libpq5     wget     gnupg     libgl1-mesa-glx     libglib2.0-0     antiword     supervisor     && r  9.9s
------
 > [docreader builder  3/13] RUN apt-get update && apt-get install -y     gcc     python3-dev     libjpeg-dev     zlib1g-dev     libpq-dev     libffi-dev     libgl1-mesa-glx     libglib2.0-0     wget     antiword     curl     unzip     && rm -rf /var/lib/apt/lists/*:
0.636 Hit:1 http://deb.debian.org/debian trixie InRelease
0.663 Hit:2 http://deb.debian.org/debian trixie-updates InRelease
0.731 Hit:3 http://deb.debian.org/debian-security trixie-security InRelease
0.802 Get:4 http://deb.debian.org/debian trixie/main amd64 Packages [9668 kB]
4.664 Get:5 http://deb.debian.org/debian trixie-updates/main amd64 Packages [2432 B]
4.729 Get:6 http://deb.debian.org/debian-security trixie-security/main amd64 Packages [5304 B]
5.871 Fetched 9675 kB in 5s (1787 kB/s)
5.871 Reading package lists...
6.840 Reading package lists...
7.713 Building dependency tree...
7.935 Reading state information...
7.965 Package libgl1-mesa-glx is not available, but is referred to by another package.
7.965 This may mean that the package is missing, has been obsoleted, or
7.965 is only available from another source
7.965
7.970 E: Package 'libgl1-mesa-glx' has no installation candidate
------
------
 > [docreader runner  3/15] RUN apt-get update && apt-get install -y     libjpeg62-turbo     libpq5     wget     gnupg     libgl1-mesa-glx     libglib2.0-0     antiword     supervisor     && rm -rf /var/lib/apt/lists/*:
0.646 Hit:1 http://deb.debian.org/debian trixie InRelease
0.689 Hit:2 http://deb.debian.org/debian trixie-updates InRelease
[...]

PR #1609, opened 2026-06-05
feat(migrations): HNSW index for 1024-dim embeddings (bge-m3)

toy0116:
## Problem

`migrations/versioned/000002_embeddings.up.sql` adds HNSW partial indexes only for `dimension = 3584` (OpenAI `text-embedding-3-large`) and `dimension = 798`. Every deployment binding **bge-m3** as the embedder — which is the default for local / on-device setups and one of the most common open-source multilingual choices — produces 1024-dim vectors and finds no matching index.

`internal/application/repository/retriever/postgres/repository.go` (around line 397-413) carefully builds its `ORDER BY embedding::halfvec(<dim>) <=> $1::halfvec(<dim>)` expression to match the indexed expression exactly so the planner picks up HNSW. With dim=1024, there is no such index, so the planner silently falls back to a sequential scan over the entire `embeddings` table.

Result: vector recall stays **correct** but adds 100-500 ms per query depending on KB size. On chat / agent paths this dominates p50 latency. Symptoms are easy to misdiagnose as embedder regression (the original report on a downstream consumer described it as "vector channel returns constant rankings" because seq-scan latency made every request indistinguishable from BM25-only timing).

## Fix

Add migration `000044_embeddings_hnsw_1024.up.sql` (with matching `.down.sql`). Same partial-index pattern as 000002 — same expression, same opclass, same `m`/`ef_construction` — just targeting `dimension = 1024`:

```sql
CREATE INDEX embeddings_embedding_idx_1024 ON embeddings
USING hnsw ((embedding::halfvec(1024)) halfvec_cosine_ops)
WITH (m = 16, ef_construction = 64)
WHERE (dimension = 1024);
```

`CONCURRENTLY` is not used because plpgsql `DO` blocks open an implicit transaction and `CREATE INDEX CONCURRENTLY` cannot run inside one — same pattern as the existing 3584/798 blocks. On fresh installs the table is empty so the cost is zero; for production backfill the operator can build the index manually with `CONCURRENTLY` first and the `IF NOT EXISTS` guard makes this migration a no-op.

## Measurement

63274-row `embeddings` table, single KB filter:

| | latency |
|---|---|
| Before (seq scan) | 100-500 ms per vector query |
| After (HNSW scan) | **27-30 ms** (`EXPLAIN ANALYZE` confirms `Index Scan using embeddings_embedding_idx_1024`) |

Planner verification:

```
Index Scan using embeddings_embedding_idx_1024 on embeddings
  Order By: ((embedding)::halfvec(1024) <=> ($0)::halfvec(1024))
  actual time=27.678..30.794 rows=10 loops=1
```

## Scope

Pure additive migration. No code changes, no behavioral changes for non-1024-dim deployments, no risk to existing 3584 / 798 HNSW paths.

### DIFF
--- a/migrations/versioned/000059_embeddings_hnsw_1024.down.sql
+++ b/migrations/versioned/000059_embeddings_hnsw_1024.down.sql
@@ -1,0 +1,2 @@
+-- Down migration: drop the 1024-dim HNSW partial index added in 000044.
+DROP INDEX IF EXISTS embeddings_embedding_idx_1024;
--- a/migrations/versioned/000059_embeddings_hnsw_1024.up.sql
+++ b/migrations/versioned/000059_embeddings_hnsw_1024.up.sql
@@ -1,0 +1,39 @@
+-- Migration 000044: HNSW index for bge-m3 / 1024-dim embeddings
+--
+-- Upstream's 000002 only created HNSW partial indexes for dim=3584 (OpenAI
+-- text-embedding-3-large) and dim=798. Both queries fall through to a
+-- sequential scan on the 1024-dim halfvec column when the bound embedder
+-- is bge-m3 — which is the default for any on-device deployment using
+-- localhub-embedder.
+--
+-- Symptom: vector recall stays "correct but slow" (seq scan of all rows in
+-- the embeddings table per query, ~100-500ms by KB size). Downstream agents
+-- that probe latency-sensitive endpoints assumed degradation.
+--
+-- Fix: same partial-index pattern, dim=1024. Matches the exact expression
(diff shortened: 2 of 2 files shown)


lyingbug:
latest migration is 57, version of migration file should be large than 57 if you want execute this migration anyway

toy0116:
@lyingbug 你说得对 · 已 rebase 到最新 main 并把 migration 重命名为 `000059_embeddings_hnsw_1024`（当前 main 上最高是 #58）· 内容不变。PR 已更新。

PR #1615, opened 2026-06-05
fix(duckdb): avoid extension install during startup

langcaiye:
## Summary
- Skip DuckDB spatial/excel extension install during server startup by default
- Add opt-in DUCKDB_INSTALL_EXTENSIONS=1 for environments that can reach DuckDB extension repository
- Prevent locked-down or offline runtimes from hanging before the HTTP listener is ready

## Why
DuckDB INSTALL may access extensions.duckdb.org. In restricted network environments the cgo call can block during startup, which prevents the server from reaching its HTTP listener and causes container probes to kill the pod. It is safer to let the service start first and only enable extension installation explicitly where egress is available.

Without the extensions, xlsx/xls-related DuckDB capabilities may fail at request time, but the service remains available.

## Testing
- git diff --check
- go test ./cmd/server
- go test ./internal/container

### DIFF
--- a/internal/container/container.go
+++ b/internal/container/container.go
@@ -1314,16 +1314,27 @@ func NewDuckDB() (*sql.DB, error) {
 		return nil, fmt.Errorf("failed to open duckdb: %w", err)
 	}
 
-	// Try to install and load required extensions.
+	// Try to install and load required extensions unless explicitly disabled.
 	//   - spatial: used for st_read_meta() to enumerate layer (sheet) names from .xlsx/.xls
 	//   - excel:   used for read_xlsx() which gives proper type inference per sheet
-	bgCtx := context.Background()
-	for _, ext := range []string{"spatial", "excel"} {
-		if _, err := sqlDB.ExecContext(bgCtx, fmt.Sprintf("INSTALL %s;", ext)); err != nil {
-			logger.Warnf(bgCtx, "[DuckDB] Failed to install %s extension: %v", ext, err)
-		}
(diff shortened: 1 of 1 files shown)


langcaiye:
部署稳定性修复

这个 PR 修复受限网络环境下服务启动可能被 DuckDB 扩展安装卡住的问题。

当前 NewDuckDB() 在启动阶段同步执行 INSTALL spatial/excel。DuckDB INSTALL 会访问 extensions.duckdb.org；在无公网出口或离线部署环境中，底层 cgo 调用可能长时间阻塞，导致 HTTP 服务无法进入监听，最终被容器探针杀掉。

调整后默认不在启动阶段安装 DuckDB 扩展，仅在显式设置 DUCKDB_INSTALL_EXTENSIONS=1 时启用。这样可以保证服务优先启动；如果运行环境确认可以访问 DuckDB 扩展仓库，再按需开启扩展安装。

lyingbug:
This change will alter the program’s default behavior. I suggest skipping the component loading logic only when it is explicitly disabled.

langcaiye:
> This change will alter the program’s default behavior. I suggest skipping the component loading logic only when it is explicitly disabled.

Thanks, agreed. I updated the PR so the default behavior is unchanged: DuckDB extensions are still installed and loaded by default. The startup skip now only happens when `DUCKDB_SKIP_EXTENSION_LOAD=1` or `true` is explicitly set for locked-down environments.

PR #1663, opened 2026-06-11
fix(frontend): make wiki citations clickable

langcaiye:
## Summary
- render Wiki chunk citation markers like `[c001]` as clickable inline reference tags
- resolve citation tags through `wiki_page.chunk_refs` and show chunk content in a drawer using the existing chunk API
- keep raw citation text when chunk refs are unavailable or out of range

## Testing
- `npm run build`


langcaiye:
example of [c001] unclickable
image

PR #1720, opened 2026-06-17
fix: clean up tenant deletion state

langcaiye:
## Summary
- soft-delete tenant membership rows when deleting a tenant
- omit dangling tenant memberships from login/current-user responses
- clear stale selected-tenant state and return to the home workspace after self-service tenant deletion
- retry once without X-Tenant-ID when the client carries a deleted tenant id

This is a follow-up to the tenant workspace deletion UI added in PR #1706. The delete entry was present, but deleting the active tenant could leave stale membership/client state behind.

## Testing
- go test ./internal/application/repository ./internal/application/service
- npm run build

### DIFF
--- a/frontend/src/stores/auth.ts
+++ b/frontend/src/stores/auth.ts
@@ -469,19 +469,6 @@ export const useAuthStore = defineStore('auth', () => {
       }
     }
 
-    if (storedSelectedTenantId) {
-      try {
-        selectedTenantId.value = Number(storedSelectedTenantId)
-        if (storedSelectedTenantName) {
-          selectedTenantName.value = storedSelectedTenantName
-        }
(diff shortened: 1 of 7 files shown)


langcaiye:
Closing this in favor of #1718. After re-reviewing the diff, #1718 already covers the core fix: cascade soft-delete tenant_members, filter stale memberships from /auth/me, and navigate to an available workspace after tenant deletion.\n\nThe remaining changes here are either overlapping or too broad for this bugfix: the global retry on `Invalid target tenant ID` may mask real tenant/permission errors, and the extra service-level test duplicates the repository/service behavior already covered by #1718. Keeping the fix in #1718 makes the change smaller and easier to review.

PR #1722, opened 2026-06-17
feat(parser): support PaddleOCR-VL API key

### LINKED ISSUE #1721: [Feature]: Support optional API Key for self-hosted PaddleOCR-VL parser
## Summary

Self-hosted PaddleOCR-VL parser configuration currently supports the service endpoint and recognition options, but it does not provide an optional API Key/token field. This makes it hard to place the PaddleOCR-VL pipeline behind an API gateway, ingress auth layer, or private service proxy that expects Bearer authentication.

## Proposal

Add an optional API Key for the self-hosted `paddleocr_vl` parser. When configured, WeKnora should send it as:

```http
Authorization: Bearer <API Key>
```

The field should be optional and preserve current behavior when empty.

## Expected behavior

- Parser Engine settings show an optional API Key field for self-hosted PaddleOCR-VL.
- Parser availability checks include the Authorization header when the key is configured.
- Actual `/layout-parsing` parse requests include the same Authorization header.
- Existing deployments without a key continue to work unchanged.

## Why this is useful

Many self-hosted PaddleOCR-VL services are deployed behind protected gateways or internal auth middleware. Supporting an optional key avoids exposing the parsing endpoint without authentication while keeping the simple endpoint-only setup working.

langcaiye:
## Summary

- add an optional `paddleocr_vl_api_key` setting for the self-hosted PaddleOCR-VL parser
- send `Authorization: Bearer <API Key>` on self-hosted PaddleOCR-VL availability checks and `/layout-parsing` parse requests
- expose the optional API Key field in Parser Engine settings and update API typings / i18n / swagger schema

Closes #1721

## Test plan

```bash
go test ./internal/infrastructure/docparser
```

`npm run type-check` was also run. It fails with pre-existing unrelated TypeScript errors that are reproducible on upstream `main` (`3b317f7e`) in files such as `src/api/agent/index.ts`, `src/api/embed/index.ts`, and `src/views/knowledge/KnowledgeBase.vue`.

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -15729,6 +15729,10 @@ const docTemplate = `{
                     "description": "PaddleOCR-VL self-hosted pipeline service (full /layout-parsing API).",
                     "type": "string"
                 },
+                "paddleocr_vl_bearer_token": {
+                    "description": "Optional Bearer token for self-hosted services protected by a gateway or reverse proxy. Official PaddleX Serving does not require this field.",
+                    "type": "string"
+                },
                 "paddleocr_vl_use_chart_recognition": {
                     "type": "boolean"
                 },
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -15722,6 +15722,10 @@
                     "description": "PaddleOCR-VL self-hosted pipeline service (full /layout-parsing API).",
                     "type": "string"
                 },
+                "paddleocr_vl_bearer_token": {
+                    "description": "Optional Bearer token for self-hosted services protected by a gateway or reverse proxy. Official PaddleX Serving does not require this field.",
+                    "type": "string"
+                },
                 "paddleocr_vl_use_chart_recognition": {
                     "type": "boolean"
                 },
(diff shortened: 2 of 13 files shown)


lyingbug:
This field is **not** part of the official PaddleOCR/PaddleX self-hosted `/layout-parsing` API. The official PaddleX serving docs only define the request body (`file`, `fileType`, etc.) and do not document any auth header.
Do you have some related docs  about it ?

langcaiye:
> This field is **not** part of the official PaddleOCR/PaddleX self-hosted `/layout-parsing` API. The official PaddleX serving docs only define the request body (`file`, `fileType`, etc.) and do not document any auth header. Do you have some related docs about it ?

Thanks for pointing this out. You are right: this is not part of the official PaddleX self-hosted `/layout-parsing` API, and the previous field name/description made that unclear.

I updated the PR to rename the setting to `paddleocr_vl_bearer_token` and clarified that it is only for deployments where the self-hosted service is protected by an external auth gateway or reverse proxy. Official PaddleX Serving does not require this field, and leaving it empty keeps the official API behavior unchanged.

I also kept backward-compatible reading of the old `paddleocr_vl_api_key` field internally, but it is no longer exposed in the public config schema.

PR #1768, opened 2026-06-23
fix: preserve queued wiki housekeeping tasks

cctv549681:
## Description
Fixes a false-positive housekeeping failure for wiki knowledge rows that are still waiting in the task queue.

This PR updates housekeeping to distinguish genuinely orphaned processing/finalizing knowledge rows from rows that are still backed by queued or active tasks. It also adds a precise pending-op lookup for wiki ingest tasks, so a finalizing wiki knowledge item is not marked failed while its durable pending operation still exists.

Changes include:
- Add read-only task queue inspection before housekeeping marks stale knowledge as failed
- Check `task_pending_ops` by exact wiki ingest dedup key for finalizing wiki rows
- Add a dedup index migration for `task_pending_ops`
- Add regression tests for queued task, wiki pending-op, and probe-error cases

## Type of Change
- [x] 🐛 Bug fix

## Related Issue
Fixes #

## Testing
```bash
go test ./internal/application/service -run 'TestHousekeeping'
go test ./internal/application/repository -run 'TestTaskPendingOps'
```

## Checklist
- [x] Self-reviewed the code

## Screenshots / Recordings
before: 

image
img_v3_0212p_fd1093bd-e473-441d-a281-f95d474ca57g

after:
image
image


lyingbug:
Thanks for investigating the housekeeping false positives. Closing this PR as superseded by #2715, together with the queued-task recovery work in #2059.

Current [filterOutQueued](https://github.com/Tencent/WeKnora/blob/b23ae055feef678c72138298f4d97693c6e49920/internal/application/service/knowledge_housekeeping.go#L291) checks durable Wiki ingest entries in `task_pending_ops` by document dedup key before consulting the transient task inspector. It preserves rows with pending Wiki work even when no inspector is available, and defers candidates when the durable lookup fails. The [current regression coverage](https://github.com/Tencent/WeKnora/blob/b23ae055feef678c72138298f4d97693c6e49920/internal/application/service/knowledge_housekeeping_test.go#L309) includes a finalizing document protected by durable Wiki work and recovery when no such work exists.

The false-failure scenario motivating this PR is covered on `main`; the alternative repository helper and migration in this historical patch are no longer needed for that fix.

PR #1791, opened 2026-06-24
fix(frontend): ensure settings modal remains usable on very small viewports

estelledc:
## Description

Follow-up to #1732 — the previous fix added `max-height: calc(100vh - 40px)` which prevents the modal from exceeding the viewport, but on very small screens (viewport height < ~440px) the modal shrinks to a point where the close button and most navigation items become unreachable.

This PR adds a minimum usable height, makes the overlay scrollable as a fallback, and uses a safe centering pattern (`margin: auto` instead of `align-items: center`) so users on small screens can always reach the close button and all nav items.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Related Issue

Refs #1731

## Root Cause

When viewport height is very small (e.g. 400px):
1. `max-height: calc(100vh - 40px)` = 360px — the modal shrinks to fit
2. `.settings-modal` has `overflow: hidden` — content beyond the modal boundary is clipped
3. `.settings-overlay` uses `align-items: center` — this is the classic flexbox overflow trap: when the modal exceeds available space, it overflows **symmetrically** (both top and bottom), making the top of the modal (including close button) unreachable even with scrolling
4. The left nav (scrollHeight: 671px) is compressed into ~300px visible area with no scroll mechanism to reach clipped items

## Changes

| File | Change |
|------|--------|
| `frontend/src/views/settings/Settings.vue` | `.settings-overlay`: add `overflow-y: auto`, change `align-items: center` → `flex-start`; `.settings-modal`: add `min-height: 400px`, `margin: auto`, `flex-shrink: 0` |

### How it works:
- **`align-items: flex-start`** on `.settings-overlay`: avoids the flexbox center + overflow trap where content overflows symmetrically, making the top unreachable
- **`margin: auto`** on `.settings-modal`: provides safe vertical centering that gracefully degrades — when space is sufficient the modal is centered; when space is tight it stays top-aligned and the overlay scrolls
- **`min-height: 400px`** on `.settings-modal`: guarantees a minimum usable area. Per CSS spec, when min-height > max-height, min-height wins — so the modal maintains 400px even if calc(100vh - 40px) would make it smaller
- **`overflow-y: auto`** on `.settings-overlay`: when the modal (400px) + padding (40px) exceeds the viewport, the overlay becomes scrollable
- **`flex-shrink: 0`**: prevents the flex container from collapsing the modal below min-height

### Before fix — viewport 250px height

The modal is compressed to 210px (`100vh - 40px`). Only 3 nav items visible (常规设置, 用户信息, API信息). Everything below is clipped with **no way to scroll or access** the remaining settings (空间信息, 成员管理, 模型管理, etc.):

before-250px

### Before fix — viewport 350px height

Slightly more visible (up to 消息管理) but still clipped at the bottom. The user **cannot scroll** the overlay to reach more content:

before-350px

### After fix — viewport 350px height

With `min-height: 400px` and scrollable overlay, the modal maintains a usable size. The user can now **scroll the overlay** to access all navigation items and settings. Note the visible scrollbar indicator on the right:

after-350px

### After fix — modal element (400px min-height enforced)

The modal itself captured via element screenshot, showing the full 400px minimum height with complete navigation visible (账户 → 空间 → 模型 sections):

after-400px-modal

## Testing

Programmatically verified at multiple viewport heights:

| Viewport | Before: overlay overflow-y | Before: modal top | After: overlay overflow-y | After: modal top | After: scrollable |
|----------|---------------------------|-------------------|---------------------------|------------------|-------------------|
| 250px | `visible` (no scroll) | clipped | `auto` | 20px | ✅ scrollTop=190 reachable |
| 350px | `visible` (no scroll) | -25px (clipped!) | `auto` | 20px | ✅ |
| 900px | N/A | centered | `auto` | 60px (centered) | Not needed |

## Checklist

- [x] Self-review of code
- [x] No regression at normal viewport sizes (900px+)
- [x] Minimal change — only CSS additions, no structural changes
- [x] Verified fix addresses the deeper align-items:center overflow trap
- [x] Tested with programmatic scroll verification

### DIFF
--- a/docreader/parser/excel_parser.py
+++ b/docreader/parser/excel_parser.py
@@ -6,6 +6,7 @@ structured Document objects with text content and chunks. It supports multiple
 sheets and handles various Excel formats using pandas.
 """
 import logging
+import re
 from io import BytesIO
 from typing import List
 
@@ -23,6 +24,22 @@ from docreader.parser.xlsx_merge import fill_merged_cells_xlsx
 from docreader.parser.xlsx_repair import repair_xlsx_bytes
 
 logger = logging.getLogger(__name__)
+
+# Pattern to detect Excel image function strings that should be excluded from
+# parsed text content.  WPS uses =DISPIMG("ID",mode) to embed images in cells;
+# when opened by other tools the formula may appear as plain text prefixed with
+# "_xlfn." or "=".  Office 365 uses =_xlfn.IMAGE(url, ...) similarly.
+# The _xlfn. prefix is optional — WPS may omit it (e.g. =DISPIMG("ID",1)).
+_IMAGE_FUNC_RE = re.compile(
+    r"^=?(_xlfn\.)?(DISPIMG|IMAGE)\(", re.IGNORECASE
+)
+
+
+def _is_image_function(value: object) -> bool:
+    """Return True if *value* looks like an embedded-image function string."""
+    if not isinstance(value, str):
+        return False
+    return _IMAGE_FUNC_RE.match(value) is not None
 
 
 class ExcelParser(BaseParser):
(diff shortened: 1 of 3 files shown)


lyingbug:
changes of docreader should not be included in this pr

PR #1833, opened 2026-06-28
feat: robustness and resumability for large-scale graph/wiki extraction

marswjf:
Improvements for running per-chunk graph extraction and wiki synthesis at
scale against heterogeneous OpenAI-compatible upstreams.

1. Adaptive response transport (models/chat/remote_api.go)
   The raw-HTTP completion path negotiates the response transport against
   what the upstream actually returns instead of assuming one shape: it
   parses a non-stream JSON response, and transparently retries as a stream
   (accumulating the SSE chunks into a full completion) when a gateway or
   channel only supports streaming, or streams even for a non-stream request.
   This keeps both non-stream-JSON and stream-only channels working without
   binding to any specific model or base URL.

2. Resume-enrichment for failed/finalizing documents
   (knowledge_resume.go, handler/router/interfaces, span tracker)
   New POST /knowledge/:id/resume-enrichment re-drives ONLY the missing
   enrichment subtasks (graph chunks + wiki) of a failed or finalizing
   document, keeping the already-produced chunks, vectors and completed
   results intact. This avoids a full reparse, which would re-chunk and
   re-embed everything just to recover an enrichment-tail failure.

3. Long-output fallback model
   (extract.go, wiki_ingest.go, wiki_ingest_batch.go, wiki_page.go)
   Optional per-KB WikiConfig.SynthesisFallbackModelID. A long-output call
   (graph extraction / wiki synthesis) that fails with a transient gateway
   error degrades to the fallback model for that single call, bounded by a
   concurrency semaphore so a rate-limited fallback model is not overwhelmed.

4. Dead-letter pending decrement (router/task.go)
   When a per-chunk graph-extract task exhausts its retries and dead-letters,
   decrement the parent knowledge's pending-subtask counter so the document
   can finalize instead of hanging in "finalizing" forever.

5. Frontend
   (DocumentListView, KnowledgeBase, KnowledgeBaseEditorModal, KBModelConfig, i18n)
   A "Resume enrichment" action on failed/finalizing documents, and a per-KB
   enrichment fallback-model selector, with zh-CN / en-US strings.


marswjf:
Superseded by #1835. This PR was missing the wiki-backlog-aware guards in the housekeeping sweep and the restart reset; without them, large batches of documents waiting in the wiki backlog are still marked `failed` under load. #1835 carries the complete, consistent set. Closing in favour of it.

PR #1835, opened 2026-06-28
feat: resilient large-scale graph/wiki extraction under a rate-limited backend

marswjf:
## Background

Building a knowledge base at scale — thousands of documents ingested concurrently while the backend LLM is rate-limited — puts per-document enrichment (per-chunk graph extraction, wiki synthesis) into a long queue that drains slowly over many hours. In that regime, several behaviours that are perfectly fine for light, sporadic ingestion turn into **batch failures**:

- A document that has finished graph extraction sits in `finalizing` for a long time waiting its turn in the wiki backlog, and the stuck-row safety nets mark it `failed` even though it is advancing normally.
- A single app restart fails the entire in-flight set at once.
- Recovering a failed document re-runs work that was already completed.
- A transient difference in how an upstream returns a response (stream vs non-stream) fails the call.

The intent of this PR is narrow and practical: **under a rate-limited backend, the ingestion queue should keep advancing robustly and complete document by document, instead of shedding large batches to `failed` and forcing a from-scratch redo.**

## Changes

**Robust queue advancement (the core of this PR)**

- **Wiki-aware housekeeping backpressure** (`knowledge_housekeeping.go`) — the stuck-row sweep skips a `finalizing` row that still has a `wiki:ingest` op queued in `task_pending_ops`. Wiki runs on that table, not asynq, so the existing asynq-only probe was blind to it and false-killed documents that were simply waiting in the wiki backlog.
- **Wiki-aware restart reset** (`container/reset_pending_tasks.go`) — on startup, rows that still have a pending `wiki:ingest` op are not marked failed; their queued work survives the restart in the DB and drains afterwards.
- **Attempt-agnostic resume** (`knowledge_resume.go`) — resume decides which graph chunks are already done by scanning every prior attempt's spans (matched by `chunk_id`), not only the latest. Restarts open fresh attempts, so completed graph spans accumulate under older attempts; the previous latest-only check re-ran the entire graph stage on every recovery.
- **Resume-enrichment endpoint** (`knowledge_resume.go`, router / handler / interfaces) — `POST /knowledge/:id/resume-enrichment` re-drives only the missing enrichment subtasks (graph chunks + wiki) of a failed/finalizing document, keeping already-produced chunks, vectors and results. Avoids a full reparse, which would re-chunk and re-embed everything just to recover an enrichment-tail failure.
- **Dead-letter pending decrement** (`router/task.go`) — when a per-chunk graph-extract task exhausts its retries and dead-letters, decrement the parent's pending-subtask counter so the document can finalize instead of hanging in `finalizing` forever.
- **Long-output fallback model** (`extract.go`, `wiki_ingest.go`, `wiki_ingest_batch.go`, `wiki_page.go`) — optional per-KB `WikiConfig.SynthesisFallbackModelID`: a long-output call (graph extraction / wiki synthesis) that fails with a transient gateway error degrades to a fallback model for that single call, bounded by a concurrency semaphore so a rate-limited fallback model is not overwhelmed.
- **Timeout-safe wiki self-chain** (`wiki_ingest_batch.go`) — `scheduleFollowUp` enqueues the next batch on the task context, which asynq cancels once a batch hits its `Timeout`. On a slow backend a batch can exhaust its retry budget on repeated timeouts; the final attempt completes its ops but the follow-up enqueue then fails on the cancelled context, so the self-chain stops and the rest of the pending queue is stranded with no recovery. The enqueue now runs on a detached context (`context.WithoutCancel`) so a single trigger keeps draining the whole backlog even when individual batches time out.

**General robustness**

- **Adaptive response transport** (`models/chat/remote_api.go`) — the raw-HTTP completion path negotiates the transport against what the upstream actually returns: it parses a non-stream JSON response and transparently retries as a stream (accumulating the SSE chunks into a full completion) when a gateway/channel only supports streaming, or streams even for a non-stream request. This keeps both non-stream-JSON and stream-only channels working without binding to any specific model or base URL.

**Frontend** (`DocumentListView`, `KnowledgeBase`, `KnowledgeBaseEditorModal`, `KBModelConfig`, i18n) — a "Resume enrichment" action on failed/finalizing documents, and a per-KB enrichment fallback-model selector, with zh-CN / en-US strings.

## Scope & impact

These changes harden the **large-scale concurrent backfill** path. They are deliberately scoped so that **light / sporadic single-document ingestion is unchanged**: the queue / watchdog / restart guards only engage when a document genuinely still has queued work (an asynq task or a `wiki:ingest` op), so on a small, normal ingest they are no-ops and behaviour is identical to today. The one change that benefits every deployment regardless of scale is the adaptive response transport, since an upstream's streaming behaviour can differ from the request in any setup.

No new required configuration (the fallback model is an optional per-KB field) and no change to default behaviour for existing knowledge bases.

## Runtime tunability (later commit)

A follow-up commit makes the concurrency knobs adjustable at runtime, so an operator can match throughput to a rate-limited backend without rebuilding or restarting:

- **Runtime-resizable fallback concurrency** — a `wiki.fallback_concurrency` system setting (default 6) drives a resizable limiter; changes apply live through the existing settings side-effect bridge (no restart). The wiki and graph fallback paths share it.
- **Fail-fast on gateway timeout** — a gateway timeout on a long-output call is deterministic, so with a fallback configured the primary degrades to it immediately rather than exhausting its retry budget on the same timeout. The primary attempt budget is itself a setting (`wiki.primary_max_attempts`, default 3).
- **Per-setting min/max** — `settingSpec` gains optional `Min`/`Max`, surfaced in the admin settings API and enforced by the number inputs.
- **Per-KB wiki build parallelism** in the editor — exposes the existing `WikiConfig` map/reduce ingest parallelism (0 = auto).
- i18n labels/descriptions for the new settings (zh-CN / en-US).

These settings all ship with sensible defaults, so existing deployments are unaffected until an operator chooses to tune them.

---

This supersedes #1833, which lacked the wiki-backlog-aware guards above. Without them, the stuck-row recovery and the restart reset still mark large batches of wiki-waiting documents `failed` under load — so #1833 on its own would regress exactly the scenario it set out to fix.


marswjf:
Closing this: after a month with no review activity, upstream has landed
equivalent fixes for part of it independently, and I've moved the rest to a
local-only patch set on top of v0.7.1.

Superseded by upstream:
- The wiki self-chain fix (00a5b13) — v0.7.1 detaches the follow-up enqueue via
  wikiIngestCleanupContext(), same root cause.
- Restart-time wiki awareness — stuckKnowledgeParseQuery() now excludes
  finalizing rows that still hold a pending wiki:ingest op.

Still absent upstream, kept locally: the resume-enrichment endpoint, the
long-output fallback model, the runtime-tunable enrichment settings, and the
housekeeping sweep's wiki-backlog blind spot (filterOutQueued still probes only
asynq, so a document waiting in the wiki backlog can be recovered as "stuck").

Happy to re-open or re-submit any individual piece if a maintainer wants it —
just say which.

PR #1855, opened 2026-06-30
feat(reparse): content-addressed stable chunk IDs (#1679 — step 1)

jianghanyun:
## What & why
Addresses step 1 of #1679. Today every re-parse deletes all chunks + vectors and recomputes everything from scratch. Because chunk IDs are random UUIDs, identical content gets a brand-new ID on every parse → 100% cache miss, so VLM OCR/Caption, embeddings and wiki extraction are all redone even when the content is unchanged.

This PR derives a **deterministic, content-addressed chunk ID** so unchanged content keeps a stable ID across re-parses. That is the prerequisite for reusing already-computed embeddings instead of recomputing them.

## Changes
- `types.NormalizeForHash` / `ContentHash` / `StableChunkID` / `ChunkIDAllocator` — content-addressed IDs (32 hex chars, fits `id VARCHAR(36)`), occurrence-disambiguated so duplicate content within one document doesn't collide on the primary key. `ContentHash` (64 hex) is populated for diffing.
- `types.PlanChunkReuse` — a pure, unit-tested diff planner that classifies incoming chunks into reuse / re-embed / delete sets. Mirrors the existing FAQ-import reuse logic (`knowledge_faq_import.go`).
- `processChunks` now sets stable ID + `ContentHash` for parent and text chunks.
- Unit tests for normalization, hashing, ID determinism + occurrence, allocator re-parse reproducibility, and the reuse planner (incl. the legacy empty-hash case).

No schema change (the `content_hash` column already exists). This PR does **not** change the reparse delete/recompute behavior yet — it only makes IDs deterministic and adds the planner, so it is safe to land on its own.

## Follow-up — P2 (embedding reuse) needs a design call
The actual "avoid full recompute" win comes from not re-embedding unchanged chunks on reparse. I've prototyped the planner (`PlanChunkReuse`) but would like maintainer input on the integration before sending it, because it touches reparse delete semantics. Two options:

- **Option A — incremental reuse:** stop the upfront `cleanupKnowledgeResources` delete-all in `ReparseKnowledge`; in `processChunks`, diff against existing chunks via `PlanChunkReuse`, `CreateChunks` only new/changed chunks, `UpdateChunks` the reused rows' position metadata (keeping their vectors), and delete only removed chunks via `DeleteByChunkIDList`. Skips embedding for unchanged content. Smaller storage footprint, but changes reparse delete semantics.
- **Option B — embedding cache table:** keep the delete-all flow; add an `embedding_cache(content_hash, embedding_model_id, dim, vector)` table looked up before embedding. No change to reparse semantics, but adds a table + migrations for mysql / sqlite / paradedb / versioned and extra storage.

Which direction do you prefer? I'm happy to implement either in a follow-up PR. Embedding-model-change invalidation is handled by keying on / checking the model id in both designs.

## Testing
- `go test ./internal/types/...` → `ok`
- `go build ./internal/...` → clean

### DIFF
--- a/internal/application/repository/chunk.go
+++ b/internal/application/repository/chunk.go
@@ -445,6 +445,18 @@ func (r *chunkRepository) DeleteChunksByKnowledgeID(ctx context.Context, tenantI
 	).Delete(&types.Chunk{}).Error
 }
 
+// PurgeSoftDeletedByKnowledgeID permanently removes rows already soft-deleted
+// for a knowledge ID. Content-addressed chunk IDs (#1679) are deterministic, so
+// re-parsing unchanged content produces the same primary key as a row the
+// reparse cleanup just soft-deleted; without purging, the recreate INSERT would
+// hit a primary-key conflict. Only soft-deleted rows are touched (deleted_at IS
+// NOT NULL), so live data is never affected.
+func (r *chunkRepository) PurgeSoftDeletedByKnowledgeID(ctx context.Context, tenantID uint64, knowledgeID string) error {
+	return r.db.WithContext(ctx).Unscoped().Where(
+		"tenant_id = ? AND knowledge_id = ? AND deleted_at IS NOT NULL", tenantID, knowledgeID,
+	).Delete(&types.Chunk{}).Error
+}
+
 // ListImageInfoByKnowledgeIDs returns non-empty image_info values for the given knowledge IDs.
 // No chunk_type filter — collects from text, image_ocr, and image_caption chunks.
 func (r *chunkRepository) ListImageInfoByKnowledgeIDs(
--- a/internal/application/repository/chunk_sqlite_test.go
+++ b/internal/application/repository/chunk_sqlite_test.go
@@ -197,6 +197,59 @@ func TestCreateChunks_SQLite_SeqIDAfterSoftDelete(t *testing.T) {
 	assert.Equal(t, int64(5), saved[1].SeqID)
 }
 
+// TestPurgeSoftDeletedByKnowledgeID_AllowsStableIDReinsert reproduces the
+// content-addressed-ID reparse conflict (#1679) and verifies the purge fixes it:
+// re-parsing unchanged content yields the same primary key as a row the reparse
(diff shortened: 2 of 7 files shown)


jianghanyun:
A heads-up and a design question before I build the embedding-reuse follow-up (#1679).

**1. Soft-delete × content-addressed IDs — fixed in this PR.**
While prototyping the reparse reuse I found that making chunk IDs deterministic interacts with chunk soft-deletion: re-parsing unchanged content produces the same primary key as the row the reparse cleanup just soft-deleted, so the recreate `INSERT` would hit a PK conflict. Fixed in commit 2: `chunkRepository.PurgeSoftDeletedByKnowledgeID` (Unscoped, `deleted_at IS NOT NULL` only) called in `processChunks` right before `CreateChunks` — a no-op on the first parse. A sqlite regression test reproduces the conflict and verifies the fix. So this PR is self-contained and safe to land on its own.

**2. Direction for the actual reuse (the "avoid full recompute" goal).**
I have a pure `PlanChunkReuse` diff planner ready. The remaining question is how to integrate it, because reusing vectors means the reparse must stop deleting them up front — and `cleanupKnowledgeResources` currently bundles several things together: vector deletion, chunk-row deletion, extracted-image cleanup, graph (`DelGraph`) cleanup, and tenant storage accounting. So incremental reuse needs that path split with care. Two options:

- **Option A — incremental reuse:** split the reparse cleanup so vectors/rows for unchanged chunks survive; in `processChunks`, diff via `PlanChunkReuse`, embed only new/changed chunks, delete only removed ones (`DeleteByChunkIDList`), and recompute storage accordingly. Biggest win (covers OCR / embedding / wiki), but it touches the cleanup + storage-accounting path.
- **Option B — embedding cache table:** keep the current delete-all flow; add an `embedding_cache(content_hash, embedding_model_id, dim, vector)` table looked up before embedding. No change to reparse semantics, but adds a table + migrations (mysql / sqlite / paradedb / versioned) and stores vectors twice.

Which direction do you prefer? And if Option A, are there constraints I should respect around the storage accounting / graph cleanup in `cleanupKnowledgeResources`? Happy to implement either as a follow-up PR.

PR #1875, opened 2026-07-01
feat(oss)/oss assign signature version v1/v4

liushutan:
## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue

Fixes #

## Checklist
- [ ] Self-reviewed the code


lyingbug:
Closing this PR to consolidate the configurable OSS signature-version work into #3020: https://github.com/Tencent/WeKnora/pull/3020. That PR is still open; this is consolidation, not a claim that the feature has already shipped.

Both proposals add V1/V4 selection for the OSS client, temporary buckets, connectivity checks, and the global OSS_SIGNATURE_VERSION setting. #3020 additionally wires signature_version through the current storage-backend model, legacy conversion, connectivity test, and settings UI. It also keeps the SSRF-safe HTTP client when selecting V1; this PR rebuilds the V1 config without that transport.

Two differences from this PR remain for follow-up in #3020: V1 here also forces path-style addressing, and the per-storage factory falls back to OSS_SIGNATURE_VERSION when its field is empty. #3020 does not currently carry those behaviors, so any gateway requirement for them should be discussed and tested there explicitly.

Thank you for the contribution. Please use #3020 as the single follow-up for the signature-version implementation and these compatibility details.

PR #2010, opened 2026-07-12
feat: support MySQL as primary database

March-77:
## What changed

- add `DB_DRIVER=mysql` as a supported primary business database
- configure GORM and golang-migrate MySQL DSNs, including escaped credentials and IPv6 hosts
- add a versioned MySQL 8 schema baseline through migration version 65
- adapt PostgreSQL-specific repository queries, JSON operations, upserts, locking, and case-insensitive search for MySQL
- keep vector retrieval separate and require an external retriever when MySQL is the primary database
- add Docker Compose, Helm, environment, startup script, and documentation support
- add unit and repository-level MySQL dialect coverage
- add an opt-in integration test that migrates an empty MySQL database through the production migration entry point, verifies critical tables and version 65, then verifies idempotent restart

## Why

The project exposed `DB_DRIVER` configuration but the application startup path only accepted PostgreSQL and SQLite. Selecting MySQL therefore failed before the service could initialize, and the PostgreSQL migration set could not be applied to MySQL.

## User impact

Deployments can use MySQL 8 as WeKnora's primary relational database while continuing to use Qdrant, Milvus, Weaviate, Doris, Tencent VectorDB, Elasticsearch, or OpenSearch for retrieval.

## Validation

Validated with MySQL 8.0.46 on WSL:

- empty database migrated through `RunMigrations` from no version to version 65
- a second migration run completed as a clean no-op
- all critical schema assertions passed
- all seven MySQL repository integration scenarios passed, including JSON queries, batch updates, upserts, cleanup cutoffs, wiki search, and concurrent task claims
- `go test ./internal/database`
- `go test ./internal/application/repository`
- `go test -run TestBuildDatabaseConnectionSettings ./internal/container`
- shell syntax checks for migration, entrypoint, and development scripts
- `git diff --check`

## Attribution

The primary implementation commit is retained from Xuehao Xu (`GodBlf`) and preserves its original authorship from #2006. This branch adds independently executed MySQL 8 validation and the production migration integration test in commit `5ad7c6e1`.

Closes #1418


March-77:
Closing this draft because it was based on another participant's implementation. I will rebuild the solution independently from upstream main.

PR #2011, opened 2026-07-12
feat: support MySQL as primary database

March-77:
## What changed

- add a native GORM MySQL dialector and safe DSN construction
- bootstrap the current application schema on MySQL while adapting PostgreSQL-specific model types
- add MySQL query variants for JSON lookup, case-insensitive search, wiki search, task failure counters, and storage-provider migration
- add real MySQL integration coverage for schema idempotency and core repository operations

## Why

The configuration advertised `DB_DRIVER=mysql`, but the primary database initialization only created PostgreSQL and SQLite dialectors. PostgreSQL-only migrations and repository expressions also prevented a MySQL deployment from starting and exercising common flows.

This change makes MySQL a usable primary metadata database without changing the existing PostgreSQL retrieval/vector-store default.

## Validation

- `go test ./internal/application/repository ./internal/container`
- MySQL 8.0.46: `TestMigrateMySQLSchema`
- MySQL 8.0.46: `TestMySQLPrimaryRepositoryQueries`

Fixes #1418


March-77:
Hi maintainers, I rechecked this PR against current main (412dcc4). Although GitHub reports it clean and mergeable, the synthetic merge no longer compiles: SearchMessagesByKeyword gained an ownerID argument, and WikiLogEntry was removed. Main has also added substantial repository and schema work, including migrations 66 through 84 and new PostgreSQL-specific wiki and knowledge queries, so a correct refresh now needs broader MySQL migration and query work rather than a small conflict fix. Before I rework the branch, could you confirm whether the project still wants this MySQL-primary-database approach, and whether versioned MySQL migrations are preferred over the current AutoMigrate approach? Happy to update if this direction is still desired. Thank you!

lyingbug:
Closing this implementation after reviewing the effective merge against current main. The MySQL branch bypasses versioned migrations and relies exclusively on migrateMySQLSchema's model list, but that list omits required current tables including storage_backends (types.StorageBackend) and Wiki page revisions (types.WikiPageRevision). These are used by the current storage-backend and Wiki services, so a fresh MySQL installation cannot provide the advertised current application functionality. The integration test migrates only a subset of models and cannot establish full bootstrap coverage.

MySQL support remains valuable, and the other proposals are being retained for comparison. A replacement needs a complete schema/bootstrap and upgrade contract plus real MySQL integration coverage. This is a concrete implementation gap against today's main, not a closure merely for age or conflicts; no live MySQL database test was run.

PR #2012, opened 2026-07-12
feat: reuse content-addressed rebuild artifacts

March-77:
## What changed

- derive stable UUID-shaped chunk IDs from document ID, chunk kind, stable position, and normalized content
- reuse parse artifacts by file bytes plus parser engine/configuration
- freeze OCR and caption results by image bytes, VLM model, and prompt content
- cache embeddings by normalized text, model ID, and vector dimensions
- cache document summaries, generated questions, Wiki per-document map calls, and GraphRAG per-chunk extraction
- deliberately leave Wiki reduce uncached because it depends on live cross-document state

All artifact caches use versioned Redis keys and a bounded 30-day TTL. Provider failures are never cached. Cache access is best-effort, so Redis cache errors do not replace the underlying model/parser call.

## Why

Reparse currently deletes chunks and vectors and then assigns fresh random chunk IDs. That invalidates references and forces deterministic or content-stable work to call parsers and models again. Stable IDs preserve references, while content/model/config-addressed keys make invalidation precise: changing an embedding model only invalidates embeddings; changing an image or prompt only invalidates that VLM artifact; changing parser configuration invalidates parsing.

## Validation

- `go test ./internal/application/service ./internal/container`
- cache tests cover hits and content/config/prompt/stage invalidation for parsing, VLM, embeddings, and chat artifacts
- stable ID tests cover normalization and document/kind/position/content invalidation
- `go test ./internal/...` passes all relevant packages; existing loopback HTTP tests require mutually incompatible `SSRF_WHITELIST` settings and fail independently of this patch

Fixes #1679


March-77:
Resolved the conflicts against the latest `main` while preserving the content-addressed rebuild caches and the newer upstream LLM metadata, custom prompt, storage-backend, and resource-resolution behavior.

Validation completed:
- focused content-address/cache tests pass
- `go test ./internal/application/service ./internal/container -count=1` passes
- `go vet ./internal/application/service ./internal/container` passes
- the GitHub open-source scan passes

The unrelated Feishu logging assertion and loopback SSRF test failures from `go test ./internal/...` were reproduced on an unmodified latest `main`, so they are not introduced by this PR.

The PR is now cleanly mergeable.

PR #2111, opened 2026-07-17
fix(tracing): merge Langfuse finish metadata and export auto root spans

lyingbug:
## Summary

- Merge finish-time metadata into open-time metadata on `Trace`/`Span` `Finish` instead of discarding it, so completion fields (`outcome`, `duration_ms`, etc.) appear in Langfuse alongside correlation fields (`request_id`, `http.method`, …).
- When `StartSpan` / `StartGeneration` auto-open a root trace because none is active, hold the handle and end it on `Finish` so the root span is exported and child observations no longer have dangling parents.
- Stop calling `otel.SetTextMapPropagator` globally to avoid interfering with other OTel instrumentation in the same process.
- Add regression tests for metadata merge and auto-root export.

## Test plan

- [x] `go test ./internal/tracing/langfuse/...`

### DIFF
--- a/internal/tracing/langfuse/events.go
+++ b/internal/tracing/langfuse/events.go
@@ -23,8 +23,6 @@ const (
 	attrObsInput           = "langfuse.observation.input"
 	attrObsOutput          = "langfuse.observation.output"
 	attrObsMetadata        = "langfuse.observation.metadata"
-	attrObsLevel           = "langfuse.observation.level"
-	attrObsStatusMessage   = "langfuse.observation.status_message"
 	attrObsModel           = "langfuse.observation.model.name"
 	attrObsModelParams     = "langfuse.observation.model.parameters"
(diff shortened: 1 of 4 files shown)


bhallatina9:
> ## Summary
> 
> - Merge finish-time metadata into open-time metadata on `Trace`/`Span` `Finish` instead of discarding it, so completion fields (`outcome`, `duration_ms`, etc.) appear in Langfuse alongside correlation fields (`request_id`, `http.method`, …).
> - When `StartSpan` / `StartGeneration` auto-open a root trace because none is active, hold the handle and end it on `Finish` so the root span is exported and child observations no longer have dangling parents.
> - Stop calling `otel.SetTextMapPropagator` globally to avoid interfering with other OTel instrumentation in the same process.
> - Add regression tests for metadata merge and auto-root export.
> 
> ## Test plan
> 
> - [x] `go test ./internal/tracing/langfuse/...`

PR #2124, opened 2026-07-18
fix(client): reassemble multi-line SSE data before parsing

### LINKED ISSUE #2121: [Bug]: Go client SSE parser fails when one event contains multiple data lines
### Affected Component

Backend Service & API

### Bug Description

The Go client SSE parsers treat each `data:` line as a complete JSON payload.

For a valid Server-Sent Events frame containing multiple `data:` lines in the same event, the client overwrites the previous buffered data line and only parses the final fragment. This causes JSON parsing to fail before the stream callback is invoked.

This affects the Go client streaming paths, including agent stream, knowledge QA stream, and continue stream.

### Steps to Reproduce

1. Use commit `7807dee42a62a149c947e6e59ba3d1acea9a2c89`.

2. Add a focused regression test in `client/streaming_test.go` that sends one SSE event with multiple `data:` lines:

   ```text
   data: {"response_type":"answer",
   data: "content":"hello","done":false}
   ```

3. Run the Go client test suite:

   ```powershell
   cd client
   go test -count=1 ./...
   ```

4. The test fails while parsing the SSE event payload.

### Expected Behavior

The Go client should accumulate all `data:` lines in the same SSE event, join them with newline separators, parse the full payload once at the blank-line dispatch boundary, and invoke the stream callback normally.

For the example frame, the callback should receive:

- `response_type = answer`
- `content = hello`
- `done = false`

### Actual Behavior

The parser overwrites the previous buffered `data:` line with the last one.

It ends up trying to parse only this fragment:

```text
"content":"hello","done":false}
```

This is not valid top-level JSON, so the client returns:

```text
failed to parse SSE data: invalid character ':' after top-level value
```

The stream callback is not invoked for that event.

### WeKnora Version

Built from source at commit 7807dee42a62a149c947e6e59ba3d1acea9a2c89

### Deployment Method

Build from source

### Operating System

Windows 11

### Relevant Logs

```shell
--- FAIL: TestProcessAgentSSEStream_MultilineDataFrame (0.00s)
    streaming_test.go:51: multiline SSE data frame failed: failed to parse SSE data: invalid character ':' after top-level value
--- FAIL: TestKnowledgeQAStream_MultilineDataFrame (0.00s)
    streaming_test.go:77: multiline knowledge SSE data frame failed: failed to parse SSE data: invalid character ':' after top-level value
--- FAIL: TestContinueStream_MultilineDataFrame (0.00s)
    streaming_test.go:103: multiline continue SSE data frame failed: failed to parse SSE data: invalid character ':' after top-level value
FAIL
FAIL    github.com/Tencent/WeKnora/client
```

### Confirmation

draix:
## Summary
Per the WHATWG Server-Sent Events spec, a single event may carry multiple `data:` lines whose payloads are concatenated with a newline (`\n`) to form the final data value. All three Go client stream parsers (agent stream, knowledge QA stream, continue stream) instead overwrote the buffer on every `data:` line, so an event whose JSON payload was split across several `data:` lines was truncated to its last fragment and failed to unmarshal before the stream callback ran.

## Fix
Introduce a shared `appendSSEData` helper that strips the `data:` field name (and one optional leading space, per spec) and concatenates fragments with `\n`, and use it in all three parsers.

## Tests
Added regression tests in `client/streaming_test.go` covering the helper and all three streaming paths. They fail before this change and pass after; the full `go test ./...` suite is green.

Fixes #2121

### DIFF
--- a/client/agent.go
+++ b/client/agent.go
@@ -146,9 +146,11 @@ func (c *Client) processAgentSSEStream(reader io.Reader, callback AgentEventCall
 			continue
 		}
 
-		// Process lines with data: prefix
+		// Process lines with data: prefix. Multiple data lines within one
+		// event are concatenated with a newline (SSE spec) instead of
+		// overwriting, so events split across several data lines parse.
 		if strings.HasPrefix(line, "data:") {
-			dataBuffer = strings.TrimSpace(line[5:]) // Remove "data:" prefix
+			dataBuffer = appendSSEData(dataBuffer, line)
 		}
 	}
 
--- a/client/session.go
+++ b/client/session.go
@@ -335,9 +335,11 @@ func (c *Client) KnowledgeQAStream(
 			debugLogger.Debug("sse_event_type_set", "event_type", eventType)
 		}
 
-		// Process lines with data: prefix
+		// Process lines with data: prefix. Multiple data lines within one
+		// event are concatenated with a newline (SSE spec) instead of
+		// overwriting, so events split across several data lines parse.
 		if strings.HasPrefix(line, "data:") {
-			dataBuffer = line[5:] // Remove "data:" prefix
+			dataBuffer = appendSSEData(dataBuffer, line)
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


lyingbug:
Closing this overlapping implementation in favor of #2122. I compared all three client reader paths and ran the client test suite on #2122 merged with current main. #2122 preserves empty data-line separators and resets ContinueStream buffers at every event boundary, including ignored event types; this variant does not cover those boundaries. Issue #2121 remains open until a fix is merged. Thank you for the implementation and tests.

PR #2177, opened 2026-07-21
feat(cache): reuse content-addressed reparse artifacts

suwe12:
Closes #1679

## Summary

This PR introduces a database-backed, content-addressed artifact cache for document reparsing. It reuses deterministic and expensive intermediate results when the effective input and configuration are unchanged, rather than recomputing them every time a knowledge item is reparsed.

The change makes reparse work proportional to new or changed content for cacheable stages. It covers stable identities for content-derived artifacts and cache reuse for VLM OCR/caption, embeddings, document summaries, generated questions, per-document Wiki map results, and per-chunk graph extraction.

The cache is stored in the primary database, works in both Standard and Lite deployments, and does not introduce a Redis dependency. All cache operations are best-effort: cache failures are logged and treated as misses, so normal document processing continues.

### 1. Stable Content-Derived Identity

Content-derived chunks now use deterministic SHA-256-based IDs generated from the knowledge ID, chunk index, chunk type, and content hash. This gives unchanged reparse output stable identities and makes cached artifacts reusable across reprocessing.

FAQ and user-created chunks retain UUID-based IDs for backward compatibility. Generated question IDs are also derived from normalized question content rather than timestamps.

### 2. Artifact Cache Infrastructure

A new `artifact_caches` table stores payloads using a tenant-qualified composite cache identity:

```text
(tenant_id, cache_key, cache_type, input_hash, config_hash)
```

The implementation includes:

- GORM model, repository, interfaces, and service layer.
- PostgreSQL and SQLite up/down migrations.
- Best-effort `Get`, `Upsert`, prefix deletion, expiry-pruning support, and knowledge-level invalidation.
- Nil-safe cache service injection, allowing existing flows to continue if the cache is unavailable.

Configuration and model changes naturally produce cache misses through `config_hash`, avoiding broad invalidation for unrelated artifact types.

### 3. VLM OCR and Caption Cache

Image OCR and caption outputs are cached using image bytes, rendered prompt, model/interface identity, and configuration inputs. Cached VLM output becomes the canonical result for the same effective input, preventing model-output variation from destabilizing downstream content hashes.

Image child artifact IDs no longer depend on regenerated object-storage URLs.

### 4. Embedding Cache

Embedding vectors are cached by normalized input text, embedding model identity, and vector dimensions. Batch indexing now separates cache hits from misses and only calls the embedding model for misses.

The embedding cache uses a global (`tenant_id = 0`) namespace for eligible entries, enabling cross-document reuse of identical text while retaining model- and dimension-specific isolation.

### 5. Summary and Question Cache

Document summaries are cached from enriched effective document content plus rendered prompt, language, model, and output-length settings. A cache hit bypasses the summary LLM request.

Generated questions are likewise cached from their effective prompt inputs, model, language, instructions, and requested count, allowing unchanged reparses to reuse prior generation output.

### 6. Wiki Map and Graph Extraction Cache

The per-document Wiki map stage caches durable extract/classify/summary output, including slug updates and current document-specific Wiki state. Attempt-local tracing data is excluded from cached payloads.

Graph extraction is cached per chunk using its content and serialized effective extraction configuration.

Wiki reduce, taxonomy, and finalization stages are intentionally not cached because they depend on current cross-document state and must remain live.

### 7. Invalidation and Lifecycle Behavior

- Reparse retains cache entries so unchanged content can be reused.
- Permanent single and batch knowledge deletion invalidates relevant cache data.
- Cache entries are physically deleted during invalidation/expiry cleanup, allowing the same content address to be filled again later.
- All cache errors fall back to the existing computation path instead of interrupting ingestion.

## Scope Boundaries

This PR intentionally does not include:

- DocReader parsed-markdown or rendered-image artifact caching.
- Distributed single-flight coordination for concurrent cache misses.
- Cache retention scheduling, capacity controls, metrics, dashboards, or alerts.
- Caching of Wiki reduce/taxonomy/finalization stages.


suwe12:
Updated this PR with four bugs :

1. artifact_cache.go: added missing "time" import (was a compile error).
2. extract.go: restored error handling after extractor.Extract() — was nil-deref on graph.Node + false-success span.
3. wiki_ingest_batch.go: aligned wiki-map configHash between Lookup and Store (Store was missing docTitle → cache always missed).
4. knowledge_process.go: switched processChunks from soft delete to PurgeChunksByKnowledgeID — stable chunk IDs collided on retry.

PR #2187, opened 2026-07-22
feat(knowledge): upsert file uploads by first-line hash_id

leiyantianji:
## Summary
- Text uploads (`md` / `txt` / `csv` / `json` / `html` / `mhtml`) can declare a stable identity on the **first line**: `hash_id = <id>` (also `# hash_id = <id>`).
- Same `hash_id` + unchanged content → existing `409 duplicate_file` behavior.
- Same `hash_id` + changed content → delete-then-recreate (upsert), so docs like `告警策略.md` can be re-uploaded without spawning duplicates.
- Identity is stored in `metadata.hash_id`; API docs updated.

## Test plan
- [ ] Upload a markdown file whose first line is `hash_id = alarm-policy` → creates one knowledge entry with `metadata.hash_id`.
- [ ] Re-upload the identical file → `409 duplicate_file`, no second entry.
- [ ] Change body, keep the same `hash_id` line, re-upload → old entry replaced, retrieval reflects new content.
- [ ] Upload without `hash_id` → legacy content-hash dedupe unchanged.
- [ ] `go test ./internal/application/service/ -run 'TestExtractDocumentHashID|TestSupportsDocumentHashID|TestCreateKnowledgeFromFile'`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/api/knowledge.md
+++ b/docs/api/knowledge.md
@@ -58,6 +58,28 @@
 
 `process_config` 可选字段包括：`parser_engine_rules`、`chunking_config`、`enable_multimodel`、`vlm_config`、`asr_config`、`question_generation_config`、`graph_enabled`、`extract_config`。若同时传 `enable_multimodel` 与 `process_config.enable_multimodel`，以 `process_config` 为准。
 
+### 文档身份 `hash_id`（文本文件 upsert）
+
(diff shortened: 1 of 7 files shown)


lyingbug:
Putting `hash_id` on the first line of the file may not be a general solution.

PR #2189, opened 2026-07-22
[codex] Route inline think tags in chat stream

### LINKED ISSUE #2179: Lite: LLM think reasoning tags leak into chat responses (no stripping in chat_pipeline)
## Bug Description

When using an OpenAI-compatible LLM that emits `<think>...</think>` reasoning tags (e.g., MiniMax M3, DeepSeek-R1, Qwen3), the raw `<think>` content leaks into the chat response stream. The chat pipeline (`internal/application/service/chat_pipeline/`) does not strip these tags before sending SSE events to the client.

Only the IM module has think-tag stripping (`internal/im/think.go: StripThinkBlocks`), but the knowledge-chat / agent-chat pipeline does not.

## Reproduction

1. Configure WeKnora Lite with MiniMax M3 as LLM (OpenAI-compatible, `interface_type=openai`)
2. Create a knowledge base, upload a document, start a chat session
3. Send any query — the SSE `response_type: "answer"` events will contain raw `<think>...</think>` content before the actual answer

## Expected Behavior

The chat pipeline should strip `<think>...</think>` blocks from streamed responses (or convert them to a separate `response_type: "thinking"` event), similar to how `internal/im/think.go` handles it for IM messages.

## Actual Behavior

Raw `<think>` tags appear in the answer stream:

```
data:{"response_type":"answer","content":"<think>The user is asking about..."}
data:{"response_type":"answer","content":"</think>\n实际回答内容..."}
```

## Environment

- WeKnora Lite v0.7.0 (commit 380e371)
- LLM: MiniMax M3 (OpenAI-compatible endpoint)
- DB: SQLite (Lite mode)
- Language: en-US

## Suggested Fix

Add think-tag stripping to `chat_pipeline` output, similar to `internal/im/think.go: StripThinkBlocks`. Could reuse the existing `thinkBlockRe` regex.

wolfkill:
Fixes #2179.

## Root Cause

Some OpenAI-compatible reasoning models stream inline `<think>...</think>` blocks inside `ResponseTypeAnswer` chunks instead of using WeKnora's separate `ResponseTypeThinking` channel.

`PluginChatCompletionStream` only routed native `ResponseTypeThinking` chunks to `EventAgentThought`. Answer chunks were decoded and emitted directly as final-answer events, so inline reasoning tags could leak to the client.

## Changes

- Route `ResponseTypeAnswer` content through the existing `agent/tools.ThinkStreamSplitter`.
- Emit inline think content as `EventAgentThought`, reusing the same decoder/reference expansion path as native thinking chunks.
- Keep visible answer content on `EventAgentFinalAnswer`.
- Flush the splitter on normal channel close and context cancellation so split tag suffixes are not dropped.
- Add a regression test where `<think>` is split across multiple answer chunks.

## Validation

- RED: `CGO_LDFLAGS='-lc++' go test ./internal/application/service/chat_pipeline -run TestStreamRoutesInlineThinkTagsFromAnswerChunks -count=1` failed before the fix because no thought event contained `hidden`.
- GREEN: `CGO_LDFLAGS='-lc++' go test ./internal/application/service/chat_pipeline -run TestStreamRoutesInlineThinkTagsFromAnswerChunks -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/application/service/chat_pipeline -run TestStreamFlushesHeldAliasOnCancel -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/application/service/chat_pipeline -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/agent/tools -run 'TestThinkStreamSplitter|TestStripThinkBlocks' -count=1`
- `CGO_LDFLAGS='-lc++' go test ./internal/application/service/chat_pipeline ./internal/agent/tools -count=1`
- `git diff --check`

### DIFF
--- a/internal/application/service/chat_pipeline/chat_completion_stream.go
+++ b/internal/application/service/chat_pipeline/chat_completion_stream.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 
+	agenttools "github.com/Tencent/WeKnora/internal/agent/tools"
 	"github.com/Tencent/WeKnora/internal/event"
 	"github.com/Tencent/WeKnora/internal/llmreference"
 	"github.com/Tencent/WeKnora/internal/llmresource"
@@ -117,6 +118,7 @@ func (p *PluginChatCompletionStream) OnEvent(ctx context.Context,
 		thinkingID := fmt.Sprintf("%s-thinking", uuid.New().String()[:8])
 		answerID := fmt.Sprintf("%s-answer", uuid.New().String()[:8])
 		thinkingOpen := false
+		inlineThinkSplitter := agenttools.NewThinkStreamSplitter()
 
 		closeThinking := func() {
 			if !thinkingOpen {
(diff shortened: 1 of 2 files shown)


lyingbug:
Thanks for the initial inline-think routing fix. Closing this PR to consolidate the same chat-stream change in #3102, which remains open for review.

Both changes route inline <think> content from answer chunks into thought events. #3102 carries this onto the current model-context path and also addresses multiple blocks, legacy history rendering, and literal tag examples inside code regions across frontend and backend.

While reviewing #3102, retain the useful teardown behavior from this proposal: any buffered splitter/decoder suffix must be handled when a stream closes or is cancelled, including providers that do not send a final Done chunk. This closure does not mean the bug is fixed on main; follow-up work should continue in #3102 rather than maintain two competing patches to the same stream loop.

PR #2219, opened 2026-07-23
feat(reparse): content-addressed stable chunk IDs (#1679 — step 1)

anhyxdy:
What & Why
This PR implements step 1 of #1679: content-addressed stable chunk IDs.
Currently, every reparse deletes and recreates chunks with new random UUIDs. This causes 100% cache misses for embeddings, VLM results, Wiki maps, etc., even when content is unchanged.
This PR introduces deterministic, content-based chunk IDs for document-derived parent/text chunks, while keeping FAQ, manual, and clone chunks on existing UUID behavior. It also adds ContentHash and soft-delete purge handling to prevent PK conflicts.
This change is a pure foundation — it does not alter the current "delete all then recreate" reparse semantics.
Changes

New pure helpers in internal/types/: NormalizeForHash, ContentHash, StableChunkID, ChunkIDAllocator
Replace uuid.New() only for parser-generated chunks
Populate content_hash field
Add PurgeSoftDeletedByKnowledgeID to repository
Unit tests + SQLite regression test for soft-delete case

Testing

go test ./internal/types -run TestChunkID
go test ./internal/application/repository -run TestChunk
Manual reparse stability verification

Follow-up
Next PRs will build cache layers (VLM, Embedding, Wiki Map, etc.) on top of these stable IDs.

### DIFF
--- a/internal/application/repository/chunk.go
+++ b/internal/application/repository/chunk.go
@@ -467,6 +467,14 @@ func (r *chunkRepository) DeleteChunksByKnowledgeID(ctx context.Context, tenantI
 	).Delete(&types.Chunk{}).Error
 }
 
+// PurgeSoftDeletedByKnowledgeID permanently removes chunks that were already
+// soft-deleted for the given knowledge ID. Live rows are intentionally excluded.
+func (r *chunkRepository) PurgeSoftDeletedByKnowledgeID(ctx context.Context, tenantID uint64, knowledgeID string) error {
+	return r.db.WithContext(ctx).Unscoped().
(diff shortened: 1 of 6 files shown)


anhyxdy:
Closing in favor of the consolidated #1679 PR: #2242.

This change is not dropped; it is preserved as commit 1 in #2242:
`feat(reparse): content-addressed stable chunk IDs (step 1 of #1679)`

I consolidated the follow-up cache layers into one PR so reviewers can see the complete reparse-cache design in a single place, while still reviewing each step by commit.

PR #2221, opened 2026-07-23
feat(reparse): cache embedding results by content hash (#1679 - step 2)

anhyxdy:
## Description
This PR adds the first reusable artifact cache for #1679: a minimal DB-backed embedding cache.

After PR #2219 (stable chunk IDs), reparse still deletes and recreates vector indexes and calls the embedding model for every unchanged chunk. This PR introduces a `content_caches` table and wraps the embedding model so that identical `(tenant_id, normalized text, model, dimensions)` inputs can skip the embedding API on reparse.

**Scope is intentionally minimal:**
- Only adds the common cache table and repository needed by future steps
- Only caches embeddings (no VLM, Wiki, Graph, or parser artifact caching)
- Does not change reparse cleanup semantics or vector index deletion
- Does not add cache cleanup on knowledge / KB delete
- Does not update fresh-install init SQL (only versioned + sqlite migrations)

## Key Changes
- **content_caches table** with unique key `(tenant_id, cache_kind, cache_key)`
- **ContentCacheRepository** interface and implementation (`GetByKey`, `Upsert`)
- **Embedding cache wrapper** (`withEmbeddingCache`) that reuses PR1's `types.ContentHash`
- **Batch deduplication**: duplicate texts in one batch only call the real embedder once
- **Best-effort**: cache read/write errors are logged and treated as misses; indexing always succeeds
- **Wired only in `processChunks`** for the main document text chunk indexing path

## Type of Change
- [x] ✨ New feature
- [x] ⚡ Performance improvement

## Related Issue
Closes #1679 (step 2, depends on #2219)

### Repository tests (SQLite)
- miss, upsert/get, overwrite, tenant isolation

### Embedding wrapper tests
- batch duplicate text calls real embedder once
- second identical batch is entirely served from cache
- changed text / model ID / dimensions trigger cache miss
- cache read/write errors do not fail `BatchEmbed`

### Verification commands
```bash
gofmt -l internal/...  # only touched Go files
go test ./internal/application/repository -run ContentCache
go test ./internal/application/service -run EmbeddingCache
git diff --check

### DIFF
--- a/internal/application/repository/content_cache.go
+++ b/internal/application/repository/content_cache.go
@@ -1,0 +1,52 @@
+package repository
+
+import (
+	"context"
+	"errors"
+	"time"
+
+	"github.com/Tencent/WeKnora/internal/types"
+	"github.com/Tencent/WeKnora/internal/types/interfaces"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+type contentCacheRepository struct {
+	db *gorm.DB
+}
+
+func NewContentCacheRepository(db *gorm.DB) interfaces.ContentCacheRepository {
+	return &contentCacheRepository{db: db}
+}
+
+func (r *contentCacheRepository) GetByKey(
+	ctx context.Context,
+	tenantID uint64,
+	cacheKind, cacheKey string,
+) (*types.ContentCacheEntry, error) {
+	var entry types.ContentCacheEntry
+	if err := r.db.WithContext(ctx).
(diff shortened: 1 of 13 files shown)


anhyxdy:
Closing in favor of the consolidated #1679 PR: #2242.

This change is not dropped; it is preserved as commit 2 in #2242:
`feat(reparse): cache embedding results by content hash (#1679 - step 2)`

I consolidated the follow-up cache layers into one PR so reviewers can see the complete reparse-cache design in a single place, while still reviewing each step by commit.

PR #2252, opened 2026-07-23
fix(agent): honor suggested-questions limit parameter

### LINKED ISSUE #2238: [Bug]: suggested-questions?limit=6   接口linit参数没启作用
### Affected Component

Backend Service & API

### Bug Description

接口linit参数没启作用，不管传几都是6

### Steps to Reproduce

接口linit参数没启作用，不管传几都是6

### Expected Behavior

接口linit受参数影响

### Actual Behavior

接口linit受参数影响

### WeKnora Version

0.6.3

### Deployment Method

Docker Compose

### Operating System

window11

### Relevant Logs

```shell

```

### Confirmation

jack-wang-176:
## Description

`GET /agents/{id}/suggested-questions?limit=N` (and the embed equivalent `GET /embed/{id}/suggested-questions`) ignored the `limit` parameter — the response always contained 6 questions regardless of the value passed (issue #2238).

Root cause: the service clamped the caller's `limit` **down** to the agent's configured `Starters.Count` (default `6`), so any `limit >= 6` collapsed to 6 and the parameter appeared to have no effect.

This PR makes `limit` the authoritative upper bound and treats `Starters.Count` as the default that applies only when the caller omits `limit`:

- **Service** (`custom_agent.go`): `Starters.Count` is used only when no explicit limit is provided. An explicit `limit` is honored and bounded by a new `suggestionMaxLimit` (30) so it cannot inflate the downstream candidate pool (`limit*5`) without bound.
- **Handlers** (`custom_agent.go`, `embed_channel.go`): forward `0` ("unspecified") when the query parameter is absent so the agent-configured count applies, instead of hardcoding `6`. The embed cap of 12 is preserved for explicit values.
- **Embed service** (`embed_channel.go`): forwards a non-positive limit through as "unspecified" instead of overriding it with 6.
- **Frontend** (`chat`, `creatChat`, embed widget + API helpers): stop hardcoding `limit=6`; the parameter is omitted so the agent-author's configured starter count is respected. An explicit count can still be requested.
- **Swagger**: updated the `limit` annotation to describe the new default/max semantics.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue
Fixes #2238
Closes #2238

## Testing

Added `internal/application/service/custom_agent_suggestion_limit_test.go` covering:
- explicit limit **above** `Starters.Count` is honored (regression for #2238);
- explicit limit **below** `Starters.Count` is honored;
- unspecified limit falls back to the agent's configured count;
- oversized limit is bounded by `suggestionMaxLimit`.

Confirmed the regression test **fails** against the old clamp (returns 6) and **passes** with the fix (returns 10).

Also ran:
- `go build ./...` — clean
- `go vet ./internal/application/service/ ./internal/handler/` — clean
- `go test ./internal/application/service/ ./internal/handler/` — suggested-questions service & handler tests pass (one unrelated handler test, `TestPutTenantParserConfigAdminPreservesRedactedSecrets`, fails on clean `main` too and is not touched by this change)
- `gofmt -l` on all changed Go files — clean

## Checklist
- [x] Self-reviewed the code
- [x] Updated related documentation (Swagger annotation)
- [ ] `make fmt && make lint && make test`: `go build`, `go vet`, `gofmt`, and the affected package tests pass locally; `golangci-lint` is not installed in this environment so `make lint` was not run.

### DIFF
--- a/frontend/src/api/embed/index.ts
+++ b/frontend/src/api/embed/index.ts
@@ -225,9 +225,11 @@ export async function getEmbedChunkById(channelId: string, token: string, chunkI
   )
 }
 
-export async function getEmbedSuggestedQuestions(channelId: string, token: string, limit = 6) {
+export async function getEmbedSuggestedQuestions(channelId: string, token: string, limit?: number) {
+  // Omit limit to let the channel agent's configured starter count apply.
+  const qs = typeof limit === 'number' && limit > 0 ? `?limit=${limit}` : ''
   return get<{ success: boolean; data: { questions: SuggestedQuestion[] } }>(
-    `/api/v1/embed/${channelId}/suggested-questions?limit=${limit}`,
+    `/api/v1/embed/${channelId}/suggested-questions${qs}`,
     { headers: { Authorization: `Embed ${token}` } },
   )
 }
--- a/frontend/src/stores/settings.ts
+++ b/frontend/src/stores/settings.ts
@@ -418,8 +418,12 @@ export const useSettingsStore = defineStore("settings", {
       return this.settings.selectedFiles || [];
     },
 
-    /** Scope for suggested-questions API (KB / file / tag @mentions). */
-    getSuggestedQuestionsParams(limit = 6) {
+    /**
+     * Scope for suggested-questions API (KB / file / tag @mentions).
+     * Leave `limit` undefined to let the backend apply the agent's configured
+     * starter count; pass a number only to request a specific count.
+     */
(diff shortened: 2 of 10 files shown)


jack-wang-176:
Superseded by a re-opened PR from the same branch with a cleaned-up commit history.

PR #2259, opened 2026-07-23
fix(chatpipeline): normalize graph triple schema aliases

### LINKED ISSUE #2256: [Bug]: Graph extraction silently drops valid JSON triples using common schema aliases
### Affected Component

Backend Service & API

### Bug Description

The graph extraction parser can successfully parse the LLM response as JSON, but then silently drops many valid graph triples because `ParseGraph` only accepts WeKnora's narrow canonical keys:

- node: `entity`, optional `entity_attributes`
- relation: `entity1`, `entity2`, `relation`

In practice, LLMs often return equivalent and common graph schemas such as:

- `subject` / `predicate` / `object`
- `head_entity` / `relation_type` / `tail_entity`
- `head` / `relation` / `tail`
- `source` / `type` / `target`
- nested `head` / `tail` objects with `text`, `type`, `attrs`

These responses are valid JSON and contain useful graph information, but they hit:

```text
Unsupported graph group
```

and are skipped. The graph chunk task is still marked `done`, so the failure is easy to miss: the pipeline consumes LLM tokens but writes zero graph nodes and zero relations.

This is different from #1113 / #1131. Those fixed JSON extraction from markdown fences or truncated responses. Here, JSON parsing succeeds; the loss happens after parsing because schema aliases are not normalized.

### Steps to Reproduce

1. Run WeKnora from source or Docker Compose with Neo4j enabled:
   ```text
   NEO4J_ENABLE=true
   ```
2. Create or use a knowledge base with graph extraction enabled.
3. Use a minimal graph config such as:
   ```json
   {
     "enabled": true
   }
   ```
   or use a weakly constrained graph extraction prompt.
4. Upload and parse a document whose chunks contain clear entity-relation facts.
5. Inspect app logs and graph-processing span output.

Observed production evidence from a Docker Compose deployment:

```sql
-- Last 30 minutes of graph chunk spans
status   | count
---------+------
done     | 33
running  | 5
failed   | 0

-- Completed graph chunk spans in the same window
spans | zero_done | nodes | rels
------+-----------+-------+-----
38    | 33        | 0     | 0
```

Example skipped groups from logs:

```text
Unsupported graph group: map[
  head_entity:Patients (Cohort of 4)
  relationship_type:Patient_Treatment
  tail_entity:Antimicrobial De-escalation Therapy
  related_attributes:Based on negative result of third ddPCR assay
]

Unsupported graph group: map[
  subject:28-day survivors
  predicate:had downward trend
  object:pathogen DNA load
  relationship_attributes:map[timing:first three ddPCR assay]
]

Unsupported graph group: map[
  head:map[id:E1 text:blood culture type:Test]
  relation:Comparison
  tail:map[id:E2 text:ddPCR type:Test]
  attrs:map[comparison_detail:...]
]
```

### Expected Behavior

Graph extraction should not silently discard valid JSON triples that use common equivalent field names.

At minimum:

1. `ParseGraph` should normalize common aliases into the existing canonical internal format:
   - `subject` / `object` -> `entity1` / `entity2`
   - `predicate`, `relation_type`, `relationship_type`, `type` -> `relation`
   - `head_entity` / `tail_entity` -> `entity1` / `entity2`
   - `source` / `target` -> `entity1` / `entity2`
   - nested `head` / `tail` objects should use `text`, `name`, or `id` as endpoint names when present
2. Existing canonical outputs must keep working unchanged.
3. If graph extraction is enabled but the effective `extract_config` has no relation tags or examples, the system should either:
   - fall back to the default graph extraction schema/examples, or
   - surface a clear configuration warning instead of running many LLM calls that produce zero graph output.

### Actual Behavior

The graph chunk tasks complete successfully but often write zero graph data.

Recent observed numbers:

```text
graph chunk spans in last 12h: 139
done spans with 0 nodes and 0 relations: 132
non-zero done spans: 1
total nodes_added: 4
total relations_added: 3
```

Relevant code path:

- `internal/application/service/chat_pipeline/extract_entity.go`
  - `ParseGraph` only accepts `entity` or `entity1/entity2/relation`.
- `internal/application/service/extract.go`
  - graph post-processing builds a prompt from the effective `extract_config`.
- `internal/types/extract_graph.go`
  - current canonical internal graph types are `GraphNode{Name, Chunks, Attributes}` and `GraphRelation{Node1, Node2, Type}`.

The issue can happen even though the LLM response is valid JSON and semantically useful.

### WeKnora Version

Built from source / Docker Compose.

Observed on upstream main commit:

```text
8cb43473 fix(knowledge): restrict source file download to Editor and above
```

The local deployment branch has deployment documentation commits on top, but the relevant parser files match upstream main.

### Deployment Method

Docker Compose

### Operating System

Ubuntu 24.04.4 LTS, Linux aarch64

### Relevant Logs

```shell
WARNING extract_entity.go:456[ParseGraph] | Unsupported graph group: map[head_entity:Patients (Cohort of 4) related_attributes:Based on negative result of third ddPCR assay relationship_type:Patient_Treatment tail_entity:Antimicrobial De-escalation Therapy]

WARNING extract_entity.go:456[ParseGraph] | Unsupported graph group: map[subject:28-day survivors predicate:had downward trend object:pathogen DNA load relationship_attributes:map[timing:first three ddPCR assay]]

WARNING extract_entity.go:456[ParseGraph] | Unsupported graph group: map[head:map[id:E1 text:blood culture type:Test] relation:Comparison tail:map[id:E2 text:ddPCR type:Test] attrs:map[comparison_detail:...]]
```

### Related Issues / PRs Checked

- #1113 / #1131: fixed JSON recovery from markdown fences/truncated responses, but this issue happens after JSON parsing succeeds.
- #2042: graph query tool does not perform true graph traversal; related graph feature gap but not the same parser problem.
- #2190: graph/multimodal trace display fix; does not change graph extraction execution.
- #2164: Helm GraphRAG env fix; deployment/env issue, not schema normalization.

### Suggested Fix

Add a non-breaking normalization layer before the current `ParseGraph` switch.

This should not require changing public API, frontend schema, Neo4j schema, or `GraphNode` / `GraphRelation`. It can map common aliases into the existing canonical fields:

```text
subject/head/head_entity/source/from -> entity1
object/tail/tail_entity/target/to -> entity2
predicate/relation_type/relationship_type/type/rel -> relation
entity/node/name -> entity
attributes/attrs/properties/props -> entity_attributes
```

For nested endpoint objects, prefer endpoint name in this order:

```text
text -> name -> id
```

This would keep the existing internal canonical format while making graph extraction robust to common LLM triple schemas.

### Confirmation

leether:
## Description

Fixes graph extraction silently dropping valid JSON triples when the LLM returns common schema aliases instead of WeKnora's canonical graph keys.

The current parser only accepts canonical keys such as:

- `entity`
- `entity1`
- `entity2`
- `relation`

In practice, LLMs often return equivalent triple schemas such as:

- `subject` / `predicate` / `object`
- `head_entity` / `tail_entity` / `relationship_type`
- `source` / `target` / `relation_type`
- nested `head` / `tail` objects with `text`, `name`, or `id`

Before this change, those outputs were valid graph triples semantically, but were dropped as unsupported graph groups.

This PR adds a small normalization pass before graph parsing. It maps common aliases to the existing canonical internal keys and leaves already-canonical output unchanged.

No API, storage schema, Swagger annotation, frontend behavior, or user-facing configuration is changed.

## Type of Change

- [x] 🐛 Bug fix
- [x] 🧪 Test

## Related Issue

Fixes #2256

## Testing

Targeted parser tests:

```bash
go test ./internal/application/service/chat_pipeline -run "TestFormater_ParseGraph|TestIsLikelyLanguageTag" -count=1
```

Result:

```text
ok  github.com/Tencent/WeKnora/internal/application/service/chat_pipeline
```

Diff-only lint for this change:

```bash
golangci-lint run --new-from-rev=origin/main ./...
```

Result:

```text
0 issues.
```

Formatting/diff checks:

```bash
git diff --check origin/main...HEAD
gofmt -l internal/application/service/chat_pipeline/extract_entity.go internal/application/service/chat_pipeline/extract_entity_test.go
```

Result: no output.

The full template command was also attempted:

```bash
make fmt && make lint && make test
```

It currently does not pass on this checkout because `make lint` reports existing full-repo baseline issues unrelated to this PR. In the clean validation worktree, `make fmt` rewrote 59 existing files and `golangci-lint run` reported 248 existing issues across unrelated packages, so the chained command stopped at `make lint` before reaching `make test`.

A separate full-repo `make test` attempt also fails on unrelated baseline/environment-sensitive tests:

- `github.com/Tencent/WeKnora/docreader/client`: expects a local docreader gRPC service at `127.0.0.1:50051`.
- `github.com/Tencent/WeKnora/internal/datasource/connector/notion`: loopback test endpoint is rejected by SSRF validation.
- `github.com/Tencent/WeKnora/internal/infrastructure/docparser`: loopback image test endpoint is rejected by SSRF validation.

Production E2E validation was also run on three real ddPCR paper PDFs with graph extraction enabled. All completed successfully:

```text
202504 ddPCR paper: completed, graph 48/48, failed 0, Neo4j 450 nodes / 548 relations
202505 ddPCR paper: completed, graph 46/46, failed 0, Neo4j 314 nodes / 367 relations
202509 ddPCR paper: completed, graph 56/56, failed 0, Neo4j 526 nodes / 497 relations
```

Application logs during the E2E window:

```text
Unsupported graph group: 0
failed to parse graph: 0
failed to add graph: 0
panic/ERROR/FATAL: 0
```

## Checklist

- [x] Self-reviewed the code

## Screenshots / Recordings

Not applicable. This is a backend parser fix with tests.

### DIFF
--- a/internal/application/service/chat_pipeline/extract_entity.go
+++ b/internal/application/service/chat_pipeline/extract_entity.go
@@ -417,6 +417,117 @@ func (f *Formater) parseOutput(ctx context.Context, text string) ([]map[string]i
 	return itemsList, nil
 }
 
+// normalizeGroup maps common LLM output aliases to the canonical Formater
+// keys so graph extraction is robust to variations of the same triple schema
+// (e.g. RDF-style subject/predicate/object, or head_entity/tail_entity/relation)
+// instead of the canonical entity1/entity2/relation. It is a pure key-
+// normalization pass: the canonical keys themselves are in the alias lists,
+// so existing output passes through unchanged.
+func (f *Formater) normalizeGroup(group map[string]interface{}) map[string]interface{} {
+	source, hasSource := graphValueByAlias(group, f.relationSource, "subject", "head_entity", "head", "source", "from")
+	target, hasTarget := graphValueByAlias(group, f.relationTarget, "object", "tail_entity", "tail", "target", "to")
+	if hasSource && hasTarget {
+		out := map[string]interface{}{
+			f.relationSource: graphName(source),
+			f.relationTarget: graphName(target),
+		}
+		relation, ok := graphValueByAlias(
+			group,
+			f.relationPrefix,
+			"predicate",
+			"relation_type",
+			"relationship_type",
+			"type",
+			"rel",
+			"edge",
+		)
+		if ok {
+			out[f.relationPrefix] = graphName(relation)
+		}
+		return out
+	}
+
+	entity, hasEntity := graphValueByAlias(group, f.nodePrefix, "node", "subject", "object", "name", "title", "id")
+	if hasEntity {
+		out := map[string]interface{}{
+			f.nodePrefix: graphName(entity),
+		}
+		attributes, ok := graphValueByAlias(
+			group,
+			f.nodePrefix+f.attributeSuffix,
+			"attributes",
+			"node_attributes",
+			"attrs",
+			"properties",
+			"props",
+		)
+		if ok {
+			out[f.nodePrefix+f.attributeSuffix] = graphAttributes(attributes)
+		}
+		return out
+	}
+
+	return group
+}
+
+func graphValueByAlias(group map[string]interface{}, aliases ...string) (interface{}, bool) {
+	for _, alias := range aliases {
+		for key, value := range group {
+			if value != nil && strings.EqualFold(key, alias) {
+				return value, true
+			}
+		}
+	}
+	return nil, false
+}
+
+func graphName(value interface{}) string {
+	switch typed := value.(type) {
+	case map[string]interface{}:
+		if nested, ok := graphValueByAlias(typed, "text", "name", "title", "id", "label", "entity"); ok {
+			return graphName(nested)
+		}
+	case map[string]string:
+		for _, alias := range []string{"text", "name", "title", "id", "label", "entity"} {
+			for key, nested := range typed {
+				if strings.EqualFold(key, alias) {
+					return strings.TrimSpace(nested)
+				}
+			}
+		}
+	}
+	return strings.TrimSpace(fmt.Sprintf("%v", value))
+}
+
+func graphAttributes(value interface{}) []interface{} {
+	switch typed := value.(type) {
+	case nil:
+		return nil
+	case []interface{}:
+		return typed
+	case []string:
+		out := make([]interface{}, 0, len(typed))
+		for _, item := range typed {
+			out = append(out, item)
+		}
+		return out
+	case map[string]interface{}:
+		out := make([]interface{}, 0, len(typed))
+		for key, item := range typed {
+			out = append(out, fmt.Sprintf("%s: %v", key, item))
+		}
+		return out
+	case map[string]string:
+		out := make([]interface{}, 0, len(typed))
+		for key, item := range typed {
+			out = append(out, fmt.Sprintf("%s: %s", key, item))
+		}
+		return out
+	default:
+		return []interface{}{fmt.Sprintf("%v", value)}
+	}
+}
+
 func (f *Formater) ParseGraph(ctx context.Context, text string) (*types.GraphData, error) {
 	matchData, err := f.parseOutput(ctx, text)
 	if err != nil {
@@ -433,6 +544,7 @@ func (f *Formater) ParseGraph(ctx context.Context, text string) (*types.GraphDat
 	var relations []*types.GraphRelation
 
 	for _, group := range matchData {
+		group = f.normalizeGroup(group)
 		switch {
 		case group[f.nodePrefix] != nil:
 			attributes := make([]string, 0)
--- a/internal/application/service/chat_pipeline/extract_entity_test.go
+++ b/internal/application/service/chat_pipeline/extract_entity_test.go
@@ -259,3 +259,112 @@ func TestIsLikelyLanguageTag(t *testing.T) {
 		})
 	}
 }
+
+// TestFormater_ParseGraph_AliasNormalization verifies that common LLM output
+// variations of the same triple schema are normalized to the canonical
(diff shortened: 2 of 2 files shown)


internal/application/service/chat_pipeline/extract_entity.go
```diff
@@ -417,6 +417,117 @@ func (f *Formater) parseOutput(ctx context.Context, text string) ([]map[string]i
...
+	case map[string]string:
+		for _, alias := range []string{"text", "name", "title", "id", "label", "entity"} {
+			for key, nested := range typed {
+				if strings.EqualFold(key, alias) {
+					return strings.TrimSpace(nested)
+				}
+			}
+		}
+	}
+	return strings.TrimSpace(fmt.Sprintf("%v", value))
```
Lion-Leporidae, line 499: I found one edge case here: when a nested endpoint object has no recognized scalar field such as text, name, or id, graphName falls back to fmt.Sprintf("%v", value).
For example:
[{"head":{"foo":"bar"},"tail":{"text":"Bob"},"relation":"knows"}]
This currently creates a graph node named map[foo:bar]. I reproduced it with a focused test. Since the input is generated by an LLM, unexpected nested structures seem plausible, and persisting a Go map debug representation could pollute the graph.
Would it be safer for graphName to return (string, bool) and let normalizeGroup skip/log groups whose endpoints do not contain a recognized non-empty name? A regression test for unsupported and empty nested endpoints would help as well.

lyingbug:
Thanks! Normalizing alias keys in extracted triples is a best-effort fallback with limited impact, and it conflicts with #2859, which targets the root cause (JSON truncation from a too-small output budget) in the same file. Closing. Please reopen with examples if you're still seeing alias-keyed triples after that lands.

PR #2302, opened 2026-07-24
fix(agent): extend web_fetch tool timeout

### LINKED ISSUE #2298: [Bug]: web_fetch is capped by generic 60s tool timeout even when browser/HTTP fetching needs longer
### Affected Component

Backend Service & API

### Bug Description

Agent tool execution wraps every tool call with a generic 60-second timeout. This is too short for `web_fetch`, which may need to run browser rendering, HTTP fallback, and optional LLM summarization.

In production-like Docker Compose testing, multiple web_fetch attempts failed consistently at about 60 seconds, even though the broader agent request was still healthy.

### Steps to Reproduce

1. Deploy WeKnora from source with Docker Compose.
2. Enable Web Search and use an agent that can call web_fetch.
3. Ask a question that causes web_search followed by web_fetch on slower public pages.
4. Inspect `messages.agent_steps` or logs for web_fetch tool durations.

### Expected Behavior

web_fetch should have its own execution timeout budget, ideally configurable, because browser rendering and network fallback are slower than cheap local tools. The generic 60-second timeout should still apply to ordinary tools.

### Actual Behavior

web_fetch fails around 60000ms with context deadline exceeded.

Observed tool step summary before local fix:

```text
web_fetch | 60025 | false | chromedp error: chromedp run failed: context deadline exceeded; http error: Get "...": context deadline exceeded
web_fetch | 60025 | false | chromedp error: chromedp run failed: context deadline exceeded; http error: Get "...": context deadline exceeded
```

After applying a local per-tool timeout for web_fetch, the same class of E2E query completed successfully, with one web_fetch call lasting about 150036ms and returning success=true.

### WeKnora Version

Built from source, upstream main commit `8cb43473`.

### Deployment Method

Docker Compose

### Operating System

Linux self-hosted Docker Compose deployment

### Relevant Logs

```shell
web_fetch|60025|false|chromedp error: chromedp run failed: context deadline exceeded; http error: Get "...": context deadline exceeded
```

### Confirmation

- [x] I searched existing timeout issues. #1801 appears to cover LLM stream timeout, but this issue is specifically about agent tool execution timeout around web_fetch.

leether:
## Description

This PR fixes `web_fetch` being capped by the generic 60-second agent tool execution timeout.

`web_fetch` can legitimately take longer than lightweight tools because it may perform browser rendering, HTTP fallback, and optional LLM summarization. This change keeps the existing 60s timeout for ordinary tools, but gives `web_fetch` its own timeout budget.

Changes:
- keep the default generic tool execution timeout at 60s
- add a dedicated default `web_fetch` tool timeout of 180s
- add `WEKNORA_WEB_FETCH_TOOL_TIMEOUT`, accepting either Go duration values like `3m` or bare seconds like `180`
- pass the new environment variable through `docker-compose.yml`
- document the variable in `.env.example`
- add tests for per-tool timeout selection and env override parsing

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test
- [x] 🔧 Configuration / Build / CI

## Related Issue

Fixes #2298

## Testing

Passed:

```shell
go test ./internal/agent ./internal/config ./internal/application/service
```

Passed incremental lint for this PR:

```shell
golangci-lint run --new-from-rev=origin/main
```

Result:

```text
0 issues
```

Also checked:

```shell
git diff --check
```

Notes:

`make lint` now runs locally after installing `golangci-lint`, but the current upstream baseline reports unrelated existing lint issues across the repository. Those are not introduced by this PR. The PR-specific incremental lint is clean.

## Checklist

- [x] Self-reviewed the code
- [x] Updated related documentation (README, `docs/`, Swagger annotations, etc.)

## Screenshots / Recordings

N/A. Backend/runtime configuration change only.


jiahao6635:
I re-checked this against current `main` (`eeef89d`) and rebuilt a minimal local port. The direction still looks right, but `main` now has `toolExecutionTimeout(...)` for per-tool policy and gives `shell_exec` a `10m+5s` wrapper, so the update should extend that existing helper rather than reintroduce a separate `getToolExecTimeout` path.

The policy I verified is:

- ordinary tools: `60s`
- `shell_exec`: keep the existing `10m+5s` priority
- `web_fetch`: `180s` outer budget by default, with `WEKNORA_WEB_FETCH_TOOL_TIMEOUT` as a deployment-wide override

One important detail: the current fetcher already keeps its own `60s` guard around the HTTP/Chromium fetch/render stage. The outer `180s` budget does **not** relax that network/browser safety limit; it leaves room for the optional non-streaming LLM summary after the fetch stage (whose default LLM budget is `120s`). This also matches the ~150s successful E2E observation in #2298.

For the current-main port I verified the full wiring (`env -> config.Agent -> runtime AgentConfig -> runToolCall`), and added an execution-level probe confirming the actual context deadline and `ToolExecContext.ExecTimeout` for both `web_fetch` and an ordinary tool. These checks passed locally:

- `go test ./internal/agent ./internal/config ./internal/application/service ./internal/infrastructure/web_fetch ./internal/agent/tools`
- `go test -race ./internal/agent ./internal/config`
- `go vet` for the same touched packages
- `git diff --check`

Your original config/docs/tests can otherwise be retained with the conflicts resolved around the existing helper. Happy to review the refreshed version or share a more exact diff outline if useful.

PR #2383, opened 2026-07-28
fix(migration): add schema self-healing and migration repair tool

### LINKED ISSUE #2380: [Bug]: can not upload file
### Affected Component

Backend Service & API

### Bug Description

after trying uploading file , i got 
`ERROR: null value in column "custom_metadata" of relation "knowledges" violates not-null constraint (SQLSTATE 23502)
`

### Steps to Reproduce

when click for uploading a new file in knowledge base ,  the following error apear on a a pop up 

ERROR: null value in column "custom_metadata" of relation "knowledges" violates not-null constraint (SQLSTATE 23502)

( i am on dev mode)

### Expected Behavior

nitialize CustomMetadata before creating any knowledge record. The repository layer is the safest place because it covers file, URL, manual, FAQ, and other creation paths.
Update CreateKnowledge in internal/application/repository/knowledge.go:
```
func (r *knowledgeRepository) CreateKnowledge(
	ctx context.Context,
	knowledge *types.Knowledge,
) error {
	if knowledge.CustomMetadata == nil {
		knowledge.CustomMetadata = types.JSON(`{}`)
	}

	return r.db.WithContext(ctx).Create(knowledge).Error
}
```
This ensures GORM always stores an empty JSON object instead of NULL, satisfying the database constraint while preserving the expected schema. A regression test should also verify that creating knowledge without explicitly providing custom_metadata succeeds and stores {}.

### Actual Behavior

When uploading a file, WeKnora fails to create the knowledge record and returns HTTP 500:
ERROR: null value in column "custom_metadata" of relation "knowledges"
violates not-null constraint (SQLSTATE 23502)
The database migration defines custom_metadata as:
custom_metadata JSONB NOT NULL DEFAULT '{}'::JSONB
However, the file-upload path does not initialize Knowledge.CustomMetadata. GORM explicitly inserts NULL, so PostgreSQL does not apply the column default and rejects the record.

### WeKnora Version

0.7.1

### Deployment Method

Build from source

### Operating System

kali 

### Relevant Logs

```shell

```

### Confirmation

hiasky:
## Description

Adds two defense layers against a recurring class of database migration failures:

**Problem background (see #2380):** When versioned migrations are stuck in a dirty or corrupted state, critical schema columns like `knowledges.custom_metadata` are never created. This causes `SQLSTATE 42703` ("column does not exist") on every file upload, and when the column does exist but GORM sends NULL (bypassing the column default), users hit `SQLSTATE 23502` ("violates not-null constraint").

**What this PR does:**

1. **Schema self-healing (`internal/database/schema_repair.go`)**
   - On every startup, executes a set of idempotent DDL statements (`ALTER TABLE ... ADD COLUMN IF NOT EXISTS`) for critical schema invariants, regardless of whether versioned migrations have been applied.
   - Safe to run repeatedly — a cheap no-op when the schema is already correct.
   - Opt-out via `SCHEMA_REPAIR=false` env var.
   - Every entry carries a `Source` tag (e.g. `"000078"`) so error logs point to the original migration.

2. **Migration repair command (`make migrate-repair`)**
   - Detects a dirty migration state, forces to the previous clean version, then re-runs all pending migrations.
   - Uses `IF NOT EXISTS` guards so re-running is safe.

**Design rationale:** Versioned migrations (golang-migrate) remain the source of truth. The repairs are a safety net — they close the gap when migrations fail mid-way, tracking state gets corrupted, or a database is restored from a different snapshot without its `schema_migrations` table. Only columns whose absence would cause hard 500 errors are included; non-critical changes belong in versioned migrations alone.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue
Fixes #2380

## Testing
- `go build -o /dev/null ./internal/database/` — compiles cleanly.
- `go build -o /dev/null ./internal/container/` — the startup integration compiles.
- `go build -o /dev/null ./cmd/server` — full server binary compiles.
- Verified against a database whose `schema_migrations` was stuck at version 71 (dirty) with `tenant_api_keys` missing — after restart, `custom_metadata` and other missing columns were added by the repair pass without errors.
- The existing `TestCreateKnowledgeDefaultsCustomMetadataToEmptyObject` continues to pass (this covers the `BeforeCreate` NULL guard from the prior fix commit).

## Checklist
- [x] Self-reviewed the code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ help:
 	@echo "数据库:"
 	@echo "  migrate-up        执行数据库迁移"
 	@echo "  migrate-down      回滚数据库迁移"
+	@echo "  migrate-repair    修复迁移脏状态并重试"
 	@echo ""
 	@echo "开发工具:"
 	@echo "  fmt               格式化代码"
@@ -208,6 +209,9 @@ migrate-goto:
 		exit 1; \
 	fi
 	./scripts/migrate.sh goto $(version)
+
+migrate-repair:
+	./scripts/migrate.sh repair
 
 # Generate API documentation (Swagger)
 docs:
--- a/internal/container/container.go
+++ b/internal/container/container.go
@@ -703,6 +703,16 @@ func initDatabase(cfg *config.Config) (*gorm.DB, error) {
 		logger.Infof(context.Background(), "Auto-migration is disabled (AUTO_MIGRATE=false)")
 	}
 
+	// Schema self-healing: ensure critical DDL invariants are met regardless of
+	// migration tracking state. Uses IF NOT EXISTS guards so it is a cheap no-op
+	// when the schema is already up-to-date.
+	if os.Getenv("SCHEMA_REPAIR") != "false" {
+		if err := database.EnsureSchemaRepaired(db); err != nil {
+			logger.Warnf(context.Background(), "Schema repair failed: %v", err)
+			logger.Warnf(context.Background(), "Some database columns may be missing. Run migrations manually if you encounter errors.")
+		}
+	}
+
 	// Get underlying SQL DB object
 	sqlDB, err := db.DB()
 	if err != nil {
--- a/internal/database/schema_repair.go
+++ b/internal/database/schema_repair.go
@@ -1,0 +1,78 @@
+package database
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/Tencent/WeKnora/internal/logger"
+	"gorm.io/gorm"
+)
+
+// schemaRepair holds a single idempotent DDL statement that ensures a schema
+// invariant required by the current code version. Each statement MUST use
+// IF NOT EXISTS / IF EXISTS guards so it is safe to execute on every startup,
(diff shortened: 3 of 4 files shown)


lyingbug:
Closing this implementation because schema repair executes DDL outside the AUTO_MIGRATE gate. With AUTO_MIGRATE=false and SCHEMA_REPAIR unset, startup still calls EnsureSchemaRepaired and attempts ALTER TABLE statements. That changes the existing contract for externally managed schemas. The same unconditional repair also uses PostgreSQL-specific JSONB/default syntax without a dialect guard even when SQLite is selected.

Schema diagnostics and safe recovery are useful, but a replacement needs to honor the migration opt-out, use dialect-specific migrations, and validate partially applied changes before altering migration state. This conclusion comes from the startup call path and repair statements; no production database was modified.

PR #2486, opened 2026-08-01
fix(knowledgebase): clean up wiki data when deleting a knowledge base

oscarlius:
## Description

Deleting a knowledge base runs the async `ProcessKBDelete` task, which removes embeddings, chunks, physical files, graph data and the knowledge rows — but never touches the wiki tables. After a KB with an active wiki is deleted, its `wiki_pages`, `wiki_folders`, `wiki_page_issues` and `wiki_page_revisions` rows stay orphaned forever, unreachable from any UI or API surface.

This PR adds `WikiPageRepository.DeleteByKnowledgeBase` and calls it from `ProcessKBDelete`:

- **Semantics follow each table’s existing conventions**: pages, folders and issues carry `gorm.DeletedAt` and are soft-deleted (matching the per-row `Delete`/`DeleteFolder` methods); revisions have no `deleted_at` column and are hard-deleted (matching `DeleteRevisionsByPage`).
- The sweep runs in **one transaction**, is **idempotent** (safe under asynq retries), and treats an empty KB ID as a no-op so it can never degrade into a table-wide delete.
- It sits **outside** the `len(knowledgeList) > 0` block: documents deleted one by one never touch wiki data, so a KB whose documents were already removed individually can still hold live wiki pages.
- Failures are **returned for asynq to retry** (matching the knowledge-row delete semantics for pure DB rows) instead of being swallowed as warnings; the deferred queue scrub still runs on every return path.

Note: `wiki_log_entries` is intentionally not covered — the table was removed by migration `000077_remove_wiki_log` and wiki activity now flows through the audit log.

## Type of Change
- [x] 🐛 Bug fix

## Related Issue

N/A

## Testing

- New `TestDeleteByKnowledgeBaseRemovesAllWikiData` covers: soft delete of pages/folders/issues, hard delete of revisions, already-soft-deleted rows not interfering, a sibling KB left untouched, empty KB ID not deleting the whole table, and idempotency across repeated calls.
- `go test ./internal/application/repository/ -count=1` — passes
- `go test ./internal/application/service/ -count=1` — 3 failures reproduce identically on a clean `main` checkout (Windows sqlite temp-file lock flake in unrelated tests); all other tests pass
- `gofmt -l` on changed files — clean; `git diff --check` — passes
- `golangci-lint run --new-from-rev=upstream/main ./internal/...` — 0 issues
- Full `go build ./...` is blocked by pre-existing Windows toolchain issues (duckdb prebuilt static lib incompatible with the local mingw gcc; reproduced on clean `main`), unrelated to this change

## Checklist
- [x] Self-reviewed the code
- [x] Updated related documentation (method-level comments document the soft/hard delete choice)

### DIFF
--- a/internal/application/repository/wiki_page.go
+++ b/internal/application/repository/wiki_page.go
@@ -1123,6 +1123,35 @@ func (r *wikiPageRepository) DeleteByID(ctx context.Context, id string) error {
 	return nil
 }
 
+// DeleteByKnowledgeBase removes every wiki row a knowledge base owns when the
+// KB itself is deleted. Pages, folders and issues carry gorm.DeletedAt, so
+// they follow the soft-delete semantics of the per-row Delete/DeleteFolder
+// methods above; revisions have no deleted_at column, so they are hard-deleted
+// exactly like DeleteRevisionsByPage. The sweep runs in one transaction so a
+// retried KB-delete task never observes a half-cleaned wiki. An empty kbID is
+// a no-op — it must never degrade into a table-wide delete.
+func (r *wikiPageRepository) DeleteByKnowledgeBase(ctx context.Context, kbID string) error {
+	if kbID == "" {
+		return nil
+	}
+	return r.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		if err := tx.Where("knowledge_base_id = ?", kbID).
+			Delete(&types.WikiPageRevision{}).Error; err != nil {
+			return err
+		}
+		if err := tx.Where("knowledge_base_id = ?", kbID).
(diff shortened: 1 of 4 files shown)


lyingbug:
Closing as superseded by the merged #3109. Verified against main at 9d67a0eb4: `ProcessKBDelete` calls `cleanupWikiForKnowledgeBase`, which removes pages, folders, revisions and issues with tenant/KB scoping, including when earlier document cleanup fails, and returns failures for retry. The orphaned Wiki data problem addressed here is therefore already covered on main. Thank you for the implementation and tests.

Issue #2445, opened 2026-07-31
[Bug]: wiki_replace_text only replaces the first occurrence of old_text, but builtin-wiki-fixer reports issues as fully "resolved" even when duplicate occurrences remain uncorrected


ymguo0311-cmyk:
### Affected Component

Frontend UI

### Bug Description

wiki_replace_text's underlying implementation is a single strings.Replace(content, oldText, newText, 1) — Go's count 1 means only the first occurrence is replaced. When a factual error appears more than once on the same page, the builtin-wiki-fixer agent (triggered by the "Fix" button) follows its documented workflow, calls wiki_replace_text once, then calls wiki_update_issue to mark the issue resolved and reports success to the user — without checking whether other occurrences of the same wrong text remain.

internal/agent/tools/wiki_replace_text.go:
go
  existingPage.Content = strings.Replace(existingPage.Content, params.OldText, params.NewText, 1)

The fixer agent's system prompt (config/builtin_agents.yaml, workflow step 6 / constraint 6, "Exact Match for Replace") tells the model old_text must exactly match, but never mentions that a match can occur more than once, or asks the model to check for/handle multiple occurrences.

### Steps to Reproduce

 1.Seed a page with the same incorrect fact appearing twice, e.g. entity/aurora-7 contains "first released in 2015" in two separate paragraphs, when the correct year is 2019.
2. Flag the issue (contradictory_facts) via the QA/Researcher agent, or directly via wiki_flag_issue.
3. Trigger the fixer the same way the frontend "Fix" button does:
   POST /api/v1/agent-chat/{session_id}
   {"query": "Please fix the issue (ID: <issue_id>) on page [[entity/aurora-7]].",
    "agent_enabled": true, "agent_id": "builtin-wiki-fixer"}
4. Observe the tool call sequence: wiki_read_issue → wiki_read_page → wiki_read_source_doc (x2) → wiki_replace_text (called once) → wiki_update_issue (status: resolved).
5. GET /wiki/pages/entity/aurora-7 → only the first occurrence of "first released in 2015" was corrected to 2019; the second occurrence is unchanged.
6. GET /wiki/issues → status: "resolved". GET /wiki/stats → pending_issues: 0. GET /wiki/lint → no issues (lint only checks structural properties, not factual consistency).

### Expected Behavior

One of:

wiki_replace_text should replace all occurrences of old_text (or fail loudly / return a count so the caller/agent knows how many replacements were made), and/or
[...]

Issue #2444, opened 2026-07-31
[Bug]: RunLint re-flags already-archived empty_content pages, causing AutoFix to repeatedly "re-fix" and over-count the same page


ymguo0311-cmyk:
### Affected Component

Vector Database

### Bug Description

RunLint's empty_content check does not filter on page.Status. Once AutoFix archives a short page (the intended fix for LintIssueEmptyContent in wiki_lint.go), the same page still matches the empty_content rule on every subsequent RunLint call, and every subsequent AutoFix call "re-fixes" it again (setting Status = archived on an already-archived page) and re-counts it in the fixed total.

internal/application/service/wiki_lint.go — AutoFix, case LintIssueEmptyContent:

case LintIssueEmptyContent:
    if page.PageType == types.WikiPageTypeIndex || page.PageType == types.WikiPageTypeLog {
        continue
    }
    page.Status = types.WikiPageStatusArchived
    s.wikiService.UpdatePage(ctx, page)

The per-page walk in RunLint that produces the empty_content finding has no page.Status filter, so archived pages are not excluded from re-detection.

### Steps to Reproduce

1. Create a page under the empty_content threshold (<50 chars), e.g. POST /wiki/pages {"slug":"concept/short-page", "content":"Too short."}.
2. GET /wiki/lint → page flagged as empty_content, auto_fixable: true.
3. POST /wiki/auto-fix → page status becomes archived.
4. GET /wiki/lint again → the same page is still reported as an empty_content issue.
5. POST /wiki/auto-fix again → the page is "fixed" again and counted in the response's fixed total, even though nothing about it changed.

### Expected Behavior

Once a page has been archived as the empty-content fix, it should no longer surface as an open empty_content lint issue (e.g. exclude WikiPageStatusArchived from that check, or track a "fix applied" marker), and fixed counts should not grow indefinitely for a no-op re-application.

### Actual Behavior

The page is re-detected and "re-fixed" on every lint/auto-fix cycle indefinitely, inflating fix counts and making lint reports unreliable as a signal of remaining work.

### WeKnora Version

083e9738f61c685574b4f93cec32c86adc5d9907

### Deployment Method

Docker Compose

### Operating System

macOS(Darwin 24.6.0)

### Relevant Logs

```shell

```

### Confirmation

- [x] I have searched existing issues and confirmed this is a new one

PR #2515, opened 2026-08-04
fix: replace all exact wiki text matches

quan020406:
## Description

`wiki_replace_text` currently uses `strings.Replace(..., 1)`, so a page containing repeated exact text updates only the first occurrence.

This PR:

- counts all non-overlapping exact matches with `strings.Count`;
- replaces all matches with `strings.ReplaceAll`;
- returns `replacement_count` in the tool result;
- preserves the existing exact-match, `source_refs`, not-found, and persistence behavior;
- adds regression coverage for repeated, single, missing, deletion, Unicode, Markdown, and non-recursive replacement cases.

## Testing

Validated in Linux Docker with Go 1.26:

- `go test ./internal/agent/tools -count=1`
- `go test ./internal/agent/tools -run '^TestWikiReplaceText' -count=1 -v`
- `go test ./internal/agent/tools -run '^(TestWiki|TestResolveUniqueWikiPage)' -count=1`
- `go test -race ./internal/agent/tools -run '^TestWikiReplaceText' -count=1`
- `git diff --check`

The native Windows toolchain cannot link the prebuilt DuckDB static library (`__emutls_v...`), so the complete package validation was run in Linux Docker.

### DIFF
--- a/internal/agent/tools/wiki_replace_text.go
+++ b/internal/agent/tools/wiki_replace_text.go
@@ -30,7 +30,7 @@ func NewWikiReplaceTextTool(
 	return &wikiReplaceTextTool{
 		BaseTool: NewBaseTool(
 			ToolWikiReplaceText,
-			"Replace specific exact text in a Wiki page. Ideal for minor corrections.",
+			"Replace all occurrences of specific exact text in a Wiki page. Ideal for consistent minor corrections.",
 			json.RawMessage(`{
 				"type": "object",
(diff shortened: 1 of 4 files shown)


quan020406:
This PR was superseded because the original branch also contained unrelated Issue #2444 changes. The correctly scoped PR for Issue #2445 is #2516: https://github.com/Tencent/WeKnora/pull/2516

PR #2526, opened 2026-08-04
fix(docparser): default PPT/PPTX to MarkItDown

KKtwo:
## What changed

- Default PPT and PPTX files to the MarkItDown parser when no explicit parser rule matches.
- Preserve explicit `builtin`, `mineru`, and other parser overrides.
- Keep PDF and other file types on their existing builtin default.

## Why

Presentations need a parser with reliable PPT/PPTX support. Previously an empty parser rule selected the generic builtin path instead of the already registered MarkItDown presentation parser.

## Impact

Only the no-matching-rule fallback for PPT/PPTX changes. Existing parser configuration remains authoritative.

## Checks

- Focused tests in `./internal/types` and `./internal/application/service`
- `go vet ./internal/types`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/internal/types/knowledgebase.go
+++ b/internal/types/knowledgebase.go
@@ -179,7 +179,7 @@ type ChunkingConfig struct {
 	// Separators
 	Separators []string `yaml:"separators"    json:"separators"`
 	// ParserEngineRules configures which parser engine to use for each file type.
-	// When empty, the builtin engine is used for all types.
+	// When no rule matches, presentations use MarkItDown and other types use the builtin engine.
 	ParserEngineRules []ParserEngineRule `yaml:"parser_engine_rules,omitempty" json:"parser_engine_rules,omitempty"`
 	// EnableParentChild enables two-level parent-child chunking strategy.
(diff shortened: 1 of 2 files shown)


lyingbug:
Thanks for the fix. Closing this PR as superseded by the merged parser-default work in #2892.

Current [DefaultParserEngine and ResolveParserEngine](https://github.com/Tencent/WeKnora/blob/b23ae055feef678c72138298f4d97693c6e49920/internal/types/knowledgebase.go#L284) already provide the PPT/PPTX MarkItDown fallback, normalize file type spelling, and preserve explicit parser rules. The [current tests](https://github.com/Tencent/WeKnora/blob/b23ae055feef678c72138298f4d97693c6e49920/internal/types/knowledgebase_test.go#L319) cover the presentation defaults and rule overrides. Current main additionally prefers a linked AnyDoc engine when supported, while retaining MarkItDown as the presentation fallback.

The original missing-default scenario is covered by the maintained implementation; this historical patch is no longer needed.

PR #2564, opened 2026-08-06
fix(agent): shared knowledge bases render doc_count=0 in runtime_context

ShadowSmallBaby:
## Description

`getKnowledgeBaseInfos` queries the document and FAQ listings using the tenant from the request context, but a directly shared knowledge base is owned by its *source* tenant. Since those listings are tenant-scoped, shared KBs render in `runtime_context` with `doc_count=0` and an empty `recent_docs` — the model cannot see that these KBs have any content, even though they are legitimately in scope for retrieval.

The fix resolves the query tenant per knowledge base, taking it from `SearchTargets`, which `buildSearchTargets` has already resolved and authorized for each KB (a directly shared KB carries its source tenant there):

```go
func knowledgeBaseScopesForPrompt(config *types.AgentConfig) ([]string, map[string]uint64) {
	if config == nil {
		return nil, nil
	}
	kbTenantMap := config.SearchTargets.GetKBTenantMap()
	if len(config.KnowledgeBases) > 0 {
		return config.KnowledgeBases, kbTenantMap
	}
	return config.SearchTargets.GetAllKnowledgeBaseIDs(), kbTenantMap
}
```

The tenant map must be read on **both** branches, not just the `SearchTargets` one: `KnowledgeBases` also carries shared KB IDs. `KBSelectionMode="all"` appends the results of `ListSharedKnowledgeBases` into it (`session_knowledge_qa.go:390`), and `@`-mentions flow into it through `resolveKnowledgeBases`. `SearchTargets` is the sole source only for tag-only `@`-mentions, so reading the tenant map from the `SearchTargets` branch alone would leave the common paths unfixed.

**Scope of the change:** this only widens the tenant used for the *metadata* queries (`ListFAQEntries` / `ListPagedKnowledgeByKnowledgeBaseID`). It does not change the set of knowledge bases in retrieval scope. Every KB ID reaching this point has already been resolved from the authorized agent search scope, and any KB absent from the map falls back to the request tenant (own KBs and legacy configs).

One consequence worth flagging for reviewers: passing the source tenant means the redundant `HasTenantKBPermission` check inside `ListFAQEntries` no longer trips, so the safety of this path now rests on the invariant that `SearchTarget.TenantID` is always populated by `buildSearchTargets`. This is called out in a code comment. The one hand-built `SearchTarget` outside that path (`evaluation.go:404`) leaves `TenantID` unset and therefore falls back safely.

## Type of Change
- [x] 🐛 Bug fix

## Related Issue
N/A

## Testing

Two new tests:

- `TestGetKnowledgeBaseInfos_SharedKnowledgeBaseUsesSourceTenant` — with request tenant 7 and source tenant 42, asserts the metadata query actually receives tenant 42 and that `doc_count=1`
- `TestKnowledgeBaseScopesForPrompt_ExplicitKnowledgeBasesKeepSourceTenant` — covers the primary path described above: the source tenant is still carried when `KnowledgeBases` is non-empty

```
$ gofmt -l internal/application/service/agent_service{,_test,_scope_test}.go
(no output)

$ go vet ./internal/application/service/
(passes)

$ go test ./internal/application/service/ \
    -run "TestKnowledgeBaseScopesForPrompt|TestAgentHasKnowledgeScope|TestGetKnowledgeBaseInfos|TagTarget" -count=1
ok  github.com/Tencent/WeKnora/internal/application/service  6.035s   (8/8 PASS)
```

`TestBuildAgentConfig_TagOnlyScopePreservesRetrievalTarget` was included in that run to confirm the tag-only path is not regressed by the refactor.

## Checklist
- [ ] Diff-scoped lint passes where applicable — `golangci-lint` is not installed locally, so `golangci-lint run --new-from-rev=origin/main ./...` was **not** run; needs CI coverage
- [x] Self-reviewed the code
- [ ] Updated related documentation — internal behavior fix, no public API or docs surface changed
- [x] Breaking changes are clearly called out in the description above (there are none)

### DIFF
--- a/internal/application/service/agent_service.go
+++ b/internal/application/service/agent_service.go
@@ -57,31 +57,23 @@ func agentHasKnowledgeScope(config *types.AgentConfig) bool {
 	)
 }
 
-// knowledgeBaseIDsForPrompt returns KB IDs to show in runtime_context metadata.
-// Prefer explicit KnowledgeBases; fall back to deduped IDs from SearchTargets.
-func knowledgeBaseIDsForPrompt(config *types.AgentConfig) []string {
+// knowledgeBaseScopesForPrompt returns the KB IDs to show in runtime_context
+// metadata, together with the tenant each KB should be queried under.
+//
+// The tenant map always comes from SearchTargets, which buildSearchTargets has
+// already resolved (and authorized) per KB: a directly shared KB carries its
+// source tenant there. KnowledgeBases alone cannot tell own from shared KBs —
+// both KBSelectionMode="all" and an @mention put shared KB IDs into it.
+// KBs missing from the map fall back to the caller's tenant.
+func knowledgeBaseScopesForPrompt(config *types.AgentConfig) ([]string, map[string]uint64) {
 	if config == nil {
-		return nil
+		return nil, nil
 	}
+	kbTenantMap := config.SearchTargets.GetKBTenantMap()
 	if len(config.KnowledgeBases) > 0 {
-		return config.KnowledgeBases
+		return config.KnowledgeBases, kbTenantMap
 	}
-	if len(config.SearchTargets) == 0 {
-		return nil
-	}
-	seen := make(map[string]struct{}, len(config.SearchTargets))
-	out := make([]string, 0, len(config.SearchTargets))
-	for _, target := range config.SearchTargets {
-		if target == nil || target.KnowledgeBaseID == "" {
-			continue
-		}
-		if _, ok := seen[target.KnowledgeBaseID]; ok {
-			continue
-		}
-		seen[target.KnowledgeBaseID] = struct{}{}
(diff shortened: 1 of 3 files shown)


jiahao6635 (approved):
Reviewed the source-tenant boundary at `2ce87fe`. `SearchTargets` is runtime-only (`json:"-"`) and AgentQA rebuilds it per request through `buildSearchTargets`; a foreign KB receives its source tenant only after `HasTenantKBPermission`, while the evaluation-only hand-built target leaves `TenantID=0` and safely falls back to the current tenant. The change only affects metadata queries and does not expand the KB ID set.

Verified locally:
- focused service tests pass
- focused `-race` tests pass
- `go vet ./internal/application/service` passes
- `git diff --check origin/main...HEAD` passes

The shared-KB metadata regression is covered and I found no blocking issue.

PR #2607, opened 2026-08-07
feat: allow updating API key knowledge base scope

### LINKED ISSUE #2603: [Feature]: Allow updating knowledge base scope of existing scoped API keys
### Affected Component

Backend Service & API

## Background

A scoped API key can specify `knowledge_base_ids` when it is created, but the
scope cannot be updated afterward.

Currently, changing the allowed knowledge bases requires creating a new API key,
updating the external integration, and revoking the old key.

This is inconvenient for long-running integrations such as MCP servers and
external agents, where the API key should remain stable while the tenant owner
adds or removes accessible knowledge bases.

### Proposed Solution

Add an Owner-only endpoint to update the knowledge base allow-list of an
existing tenant-scoped API key:

PUT /api/v1/tenants/:tenant_id/api-keys/:key_id/knowledge-scope

Request:

{
  "knowledge_base_ids": ["kb-id-1", "kb-id-2"]
}

Expected behavior:
Only JWT-authenticated tenant Owners can call the endpoint.
API keys cannot update their own scope.
Only active, non-full-access, tenant-scoped API keys can be updated.
All knowledge bases must exist and belong to the same tenant.
The API key token and key hash remain unchanged.
The updated scope takes effect on subsequent requests.
Shared or cross-tenant knowledge bases are out of scope for this change.

### Alternatives

Open question
Currently, an empty knowledge_base_ids array means unrestricted access to all
knowledge bases in the tenant.

Should the update endpoint preserve this behavior, or reject an empty array to
prevent accidental privilege expansion?

### Impact

My team or a small group

### Use Case

_No response_

### Additional Information

_No response_

### Confirmation

estongu:
## Description

Allow workspace owners to add or remove knowledge bases from an existing scoped API key.

- Add an API for updating `knowledge_base_ids`.
- Validate that selected knowledge bases belong to the current tenant.
- Add a knowledge base scope editor to the API integration page.
- Add backend tests for tenant isolation, scope clearing, and full-access key rejection.

## Type of Change

- [x] ✨ New feature
- [x] 🧪 Test

## Related Issue
Fixes #2603 

## Testing

- Verified adding and removing knowledge bases from an existing scoped API key.
- Verified that the updated scope takes effect immediately.
- Verified that cross-tenant knowledge bases and full-access API keys cannot be updated.
- `go test` for the changed backend packages passed.
- `npm run build` passed.

## Checklist

- [x] Targeted tests pass
- [x] Self-reviewed the code
- [x] No breaking changes

## Screenshots / Recordings

image
image


lyingbug:
If the knowledge base scope can be modified, the authorization scope should be modifiable as well. Also, this API should not be designed specifically for the knowledge base scope.

estongu:
Thanks for the clarification. I understand that the API should update the
complete authorization scope, including both capabilities and
knowledge_base_ids, instead of being specific to knowledge base scope. I
will revise the endpoint and request model accordingly.

lyingbug ***@***.***> 于2026年8月10日周一 19:39写道：

PR #2620, opened 2026-08-09
fix(knowledge): bubble embedding model resolution failure for retry

mdrkrg:
## Description

`processChunks` 在获取 embedding 模型失败时直接静默返回，`ProcessDocument` 因此将任务视为成功，knowledge 停留在 `processing` 状态

修复将错误向上传递：
- asynq 会对瞬时性故障自动重试，在最后一次重试时把 knowledge 标记为失败
- 没有重试机制的同步路径（`processDocumentFromPassage`、`triggerManualProcessing` 的 goroutine），在出错时立即标记失败

提取 `markKnowledgeFailed` / `failKnowledgeRetryable` 两个公共 helper，统一散落的失败标记逻辑

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [x] 🎨 Refactor
- [x] 🧪 Test

## Testing

- asynq 文档处理（文件导入 和 段落导入）：embedding 模型解析失败时向上返回错误，触发 asynq 自动重试，非最终重试时 knowledge 保持 `processing`
- asynq 文档处理：重试次数耗尽时 knowledge 被标记为 `failed` 并写入错误信息，不再卡在 `processing`
- 手动重新解析 asynq 任务：失败返回错误触发重试，非最终重试时 knowledge 保持 `processing`
- 手动重新解析：重试耗尽后标记 `failed` 并写入错误信息
- 同步路径（无重试机制），失败时标记 `failed` 并写入错误信息
- 异步 goroutine 路径，失败时在 goroutine 内标记 `failed` 并写入错误信息
- knowledge 已被用户取消或正在删除时，跳过失败写入（状态与错误信息不覆盖）

## Checklist
- [x] Diff-scoped lint passes where applicable (for Go: `golangci-lint run --new-from-rev=origin/main ./...`)
- [x] Self-reviewed the code

### DIFF
--- a/internal/application/service/knowledge_create.go
+++ b/internal/application/service/knowledge_create.go
@@ -1180,10 +1180,10 @@ func sanitizeManualDownloadFilename(title string) string {
 
 func (s *knowledgeService) triggerManualProcessing(ctx context.Context,
 	kb *types.KnowledgeBase, knowledge *types.Knowledge, content string, doSync bool,
-) {
+) error {
 	clean := strings.TrimSpace(content)
 	if clean == "" {
-		return
+		return nil
 	}
 
 	// Resolve embedded data:base64 images and remote http(s) images → storage, replace URLs.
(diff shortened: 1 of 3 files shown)


jiahao6635 (changes requested):
The main retry/final-state direction looks good, but the async manual-reparse path still has a context-lifetime hole.

`triggerManualProcessing` correctly detaches background work into `newCtx`, and `processChunks(newCtx, ...)` runs with it. On failure, however, the goroutine calls `markKnowledgeFailed(ctx, ...)` with the original request context. Once the move request returns or is cancelled, `markKnowledgeFailed` first calls `isKnowledgeAborted` through that cancelled context. A repository read can then return `context canceled`; `isKnowledgeAborted` deliberately treats any read error as deleting, so the terminal failed write is skipped and the row can remain stuck in `processing`/`pending`—the same class of state this PR is trying to remove.

I reproduced this with a repository fake that returns `ctx.Err()` from `GetKnowledgeByID`: the async regression times out on this head. Changing the call to `s.markKnowledgeFailed(newCtx, knowledge, err.Error())` makes it pass. Please use the detached context for the whole goroutine failure path and add a cancelled-parent-context regression test.

Other validation I ran on this head: the focused embedding-failure tests, repeated service tests, the service package, and `go vet` passed; this is the only blocking finding from my review.

lyingbug:
Thanks! main no longer leaves the row stuck in processing when the embedding model can't be resolved: `failKnowledgeOnEmbeddingModel` marks it failed with the error (#3445). Closing. If you think an unresolvable model should be retried rather than failed, please open an issue describing the scenario and we can discuss it.

PR #2641, opened 2026-08-10
fix(knowledge): populate both knowledge base counts

### LINKED ISSUE #2628: [Bug]: 无法返回切片详情
### Affected Component

Frontend UI

### Bug Description

Image

### Steps to Reproduce

Image

### Expected Behavior

Image

### Actual Behavior

Image

### WeKnora Version

a

### Deployment Method

Docker Compose

### Operating System

a

### Relevant Logs

```shell

```

### Confirmation

Lion-Leporidae:
## Description

Populate both `knowledge_count` and `chunk_count` for every knowledge base type.

Previously, document knowledge bases only queried `knowledge_count`, while FAQ
knowledge bases only queried `chunk_count`. As a result, a document knowledge
base with existing chunks could be returned with `chunk_count: 0`.

This change:

- adds a shared best-effort count enrichment helper;
- uses it for current-workspace lists, explicit-workspace lists, and single-KB details;
- keeps the two count queries independent so one failure does not suppress the other result;
- adds regression coverage for document and FAQ knowledge bases.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [x] Refactor
- [ ] Performance improvement
- [x] Test
- [ ] Configuration / Build / CI

## Related Issue

Fixes #2628

## Testing

Passed:

```text
go test ./internal/application/service -run 'Test(ListKnowledgeBasesPopulatesBothCountsForEveryType|ListKnowledgeBasesByTenantIDPopulatesBothCountsForEveryType|FillKnowledgeBaseCountsContinuesAfterKnowledgeCountFailure)$' -count=1
go test -race ./internal/application/service -run 'Test(ListKnowledgeBasesPopulatesBothCountsForEveryType|ListKnowledgeBasesByTenantIDPopulatesBothCountsForEveryType|FillKnowledgeBaseCountsContinuesAfterKnowledgeCountFailure)$' -count=1
go vet ./internal/application/service
git diff --check origin/main...HEAD
```

Broader checks:

- `go test ./internal/application/service -count=1` reaches unrelated Windows-only SQLite cleanup failures because three tests leave `weknora.db` open during `t.TempDir` cleanup. The same three failures were reproduced in a clean `origin/main` worktree.
- `go test ./... -count=1` was run. In addition to the same baseline SQLite cleanup failures, this Windows environment maps public test hosts to the restricted `198.18.0.0/15` range (triggering SSRF tests) and lacks the Unix shell behavior expected by `internal/sandbox` tests.
- `golangci-lint` is not installed in the local environment; `go vet` for the changed package passes.
- A live Docker API test was not run because Docker is unavailable in the local environment.

## Potential Impact

There are no schema, route, authorization, or response-field changes. Each
knowledge base now performs the previously omitted count query, adding one
count query per knowledge base to list/detail enrichment.

## Checklist

- [ ] Diff-scoped lint passes where applicable (local `golangci-lint` unavailable; `go vet` passes)
- [x] Full-repository checks were run, with unrelated/environment-dependent failures documented above
- [x] Self-reviewed the code
- [ ] Updated related documentation (not required; no API contract changes)
- [x] No breaking changes

## Screenshots / Recordings

Not applicable; API response statistics only.

### DIFF
--- a/internal/application/service/knowledgebase.go
+++ b/internal/application/service/knowledgebase.go
@@ -361,24 +361,7 @@ func (s *knowledgeBaseService) ListKnowledgeBases(ctx context.Context) ([]*types
 	for _, kb := range kbs {
 		kb.EnsureDefaults()
 
-		// Get knowledge count
-		switch kb.Type {
-		case types.KnowledgeBaseTypeDocument:
-			knowledgeCount, err := s.kgRepo.CountKnowledgeByKnowledgeBaseID(ctx, tenantID, kb.ID)
-			if err != nil {
-				logger.Warnf(ctx, "Failed to get knowledge count for knowledge base %s: %v", kb.ID, err)
-			} else {
-				kb.KnowledgeCount = knowledgeCount
-			}
-		case types.KnowledgeBaseTypeFAQ:
-			// Get chunk count
-			chunkCount, err := s.chunkRepo.CountChunksByKnowledgeBaseID(ctx, tenantID, kb.ID)
-			if err != nil {
-				logger.Warnf(ctx, "Failed to get chunk count for knowledge base %s: %v", kb.ID, err)
-			} else {
-				kb.ChunkCount = chunkCount
-			}
-		}
+		s.fillKnowledgeBaseResourceCounts(ctx, tenantID, kb)
 
 		// Check if there is a processing import task
 		processingCount, err := s.kgRepo.CountKnowledgeByStatus(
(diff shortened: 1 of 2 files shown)


lyingbug:
why you need chunk_count in document knowledge base ?

Lion-Leporidae:
Thanks for asking. The motivation comes directly from #2628: for a document knowledge base with 4 knowledge records and 149 stored chunks, `GET /api/v1/knowledge-bases` returns `knowledge_count: 4` but `chunk_count: 0`. Because `chunk_count` is already part of the response model, API consumers cannot distinguish “no chunks exist” from “the count was not populated.” An accurate chunk count is useful for ingestion diagnostics and API clients that need to verify that documents were actually split and indexed.

This does not change the frontend's user-facing metric: document KB cards can continue to display `knowledge_count`, while FAQ KB cards display `chunk_count`.

Code changes in this PR:

- Replaced the type-specific count branches in `ListKnowledgeBases`, `ListKnowledgeBasesByTenantID`, and `FillKnowledgeBaseCounts` with a shared `fillKnowledgeBaseResourceCounts` helper.
- The helper independently queries the existing knowledge and chunk repositories, so failure to load one count does not suppress the other.
- Added regression tests covering document and FAQ KBs, both list paths, and failure isolation.
- No database schema, route, authorization, or response-shape changes.

The trade-off is one additional `COUNT` query per knowledge base compared with the current type-specific behavior. If the intended API contract is that document KB list responses should not expose an accurate `chunk_count`, I am happy to narrow the change—for example, populate it only on the detail path or follow another scope you prefer.

PR #2714, opened 2026-08-14
fix(files): fall back when message image proxy is unavailable

leonoxo:
## Summary

Fixes protected `resource://` image rendering when a newer frontend is served by a self-hosted backend that does not yet expose the message-scoped file endpoint.

## Problem

The frontend first requests:

```text
/api/v1/sessions/:session_id/messages/:message_id/files?file_path=resource://...
```

On older compatible backends this endpoint returns `404`, and the existing hydration logic treats that as a missing file and removes the image, even when the existing authenticated `/files` endpoint can serve it.

## Change

- On a `404` from the message-scoped proxy only, retry the existing authenticated tenant endpoint:
  ```text
  /files?file_path=resource://...
  ```
- Do not apply this fallback to embed visitors, which must remain scoped to their Embed token.
- Cache the successfully hydrated Blob URL under both request keys.

## Tests

- `npx tsx --test src/utils/protectedFileAccess.test.ts src/utils/security.test.ts`
- `npm run build`
- Verified against a self-hosted v0.7.2 backend: historical RAG images render again after the message route returns `404` and `/files` succeeds.

### DIFF
--- a/frontend/src/utils/protectedFileAccess.test.ts
+++ b/frontend/src/utils/protectedFileAccess.test.ts
@@ -2,6 +2,7 @@ import assert from 'node:assert/strict'
 import test from 'node:test'
 
 import {
+  buildProtectedFileFallbackRequest,
   buildProtectedFileRequest,
   isProtectedFileProxyPath,
   isProviderFileURL,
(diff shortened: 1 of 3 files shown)


lyingbug:
Thanks! This only helps a newer frontend paired with an older backend, but the two ship together. The fallback also moves message images from per-message authorization back to tenant-wide `/files` access. Closing.

PR #2715, opened 2026-08-15
fix(wiki): stop housekeeping from force-failing documents with durable Wiki backlog

hjxccc:
## Description

The stuck-row sweep asks the asynq inspector whether a candidate still has work queued, and treats "nothing queued" as proof the row is orphaned. For Wiki that question cannot be answered by asynq at all:

- Wiki ingest work is durable **per document** in `task_pending_ops`, keyed by `dedup_key = knowledge ID` (`newWikiIngestPendingOp`, `wiki_ingest.go`).
- asynq only receives a **per-KB** wake-up trigger, whose payload carries `KnowledgeBaseID` and no knowledge ID (`enqueueWikiIngestTrigger`).
- `matchesKnowledge()` gates on `taskTypesForKnowledgeCancel` (`task_inspector.go:87`), which deliberately excludes `TypeWikiIngest` — the comment there says the set is "deliberately narrow".

So `HasQueuedTasksForKnowledge` **structurally** returns false for a document whose only outstanding work is a queued Wiki ingest. `filterOutQueued` then keeps it as a candidate and the sweep flips a perfectly healthy `finalizing` row to `failed` with `"task stuck in processing > Xh, recovered by housekeeping"`, zeroing `pending_subtasks_count`. The durable op later runs anyway.

That is how a document ends up `failed` while present in neither the dead-letter table nor any visible queue.

### Change

Probe `task_pending_ops` directly before consulting the inspector.

- A durable probe **error** defers every candidate to the next sweep rather than failing them. This is deliberately the opposite of the inspector's fail-safe direction: wrongly failing a live document is not user-recoverable, whereas waiting one more interval is. The inspector's own fail-safe behaviour is unchanged.
- A nil inspector no longer disables the whole gate, only the transient half.
- No schema change, no config, no new dependency.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test

## Related Issue

Relates to #2688 — specifically its point 4, "部分文档的 Wiki span 明确显示失败，但既不在死信队列，也不在待处理队列中，似乎被静默丢弃了". This PR addresses that silent-drop mechanism only; the other asks in that issue (403 treated as non-transient, Wiki status not surfaced at document level, Wiki-only batch retry) are out of scope here.

## Testing

Go 1.26.0, `CGO_ENABLED=1`:

- `gofmt -l internal/application/service/knowledge_housekeeping.go internal/application/service/knowledge_housekeeping_test.go` — no output
- `go vet ./internal/application/service/` — clean
- `go test ./internal/application/service/ -run Housekeeping -count=1 -v` — all 10 cases pass (8 pre-existing + 2 new)

New coverage:

- `TestHousekeeping_NoFalseKill_DurableWikiIngestPending` reproduces the bug — it **fails on the unpatched sweep** (verified by reverting only the non-test file and re-running) and passes with the fix.
- `TestHousekeeping_StillRecoversWhenNoDurableOp` guards against the new gate becoming a blanket amnesty: a stale row with no pending op of its own — including the case where a pending op exists for a *different* document — must still be recovered.

The focused package was validated. Full-repository `make test` was not run; the change is confined to the housekeeping sweep and its unit tests.

## Checklist

- [x] Targeted tests for the changed packages pass
- [x] Diff-scoped lint passes where applicable (`go vet` for the changed package)
- [x] Self-reviewed the code
- [x] Updated related documentation (not applicable; no user-facing contract or configuration change)
- [x] Breaking changes are clearly called out (none)

## Screenshots / Recordings

Not applicable; backend-only fix.

### DIFF
--- a/internal/application/service/knowledge_housekeeping.go
+++ b/internal/application/service/knowledge_housekeeping.go
@@ -42,8 +42,9 @@ type HousekeepingService struct {
 	// inspector lets the sweep distinguish a genuinely orphaned row from
 	// one whose enrichment subtasks are merely backlogged behind a busy
 	// queue (no span heartbeat yet because no worker has picked them up).
-	// nil-safe — a nil inspector disables the queue check and the sweep
-	// falls back to the span/updated_at heuristics alone.
+	// nil-safe — a nil inspector disables only the transient queue check.
+	// Durable Wiki ownership in task_pending_ops is always probed through db,
+	// so the sweep never falls back to the span/updated_at heuristics alone.
 	inspector interfaces.TaskInspector
 
 	mu      sync.Mutex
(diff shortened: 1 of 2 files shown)


hjxccc:
友好 ping 一下 🙏 CI 全绿、无冲突。

这个问题是结构性的、不是偶发：Wiki 的待办是 per-document 落在 `task_pending_ops`，而 asynq 只收到 per-KB 的唤醒触发，且 `taskTypesForKnowledgeCancel`（task_inspector.go:87）有意不含 `TypeWikiIngest`——所以 `HasQueuedTasksForKnowledge` 对这类文档必然返回 false，健康的 `finalizing` 行会被 housekeeping 直接刷成 `failed`，而它既不在死信表也不在任何可见队列里，很难被发现。我们在生产上是靠一批文档莫名 failed 才回溯到这里的。

改动只是在问 inspector 之前先探一次 `task_pending_ops`，无 schema、无配置、无新依赖。需要调整方向随时说。

PR #2779, opened 2026-08-24
fix: bound wiki reduce input growth

### LINKED ISSUE #2718: [Feature]: Wiki reduce prompt has no input budget — a hub page that outgrows the model context fails permanently and cannot recover
### Affected Component

Backend / Wiki ingest

### Problem Description

`reduceSlugUpdates` assembles the page-modify prompt from five potentially-large inputs — `ExistingContent`, `SharedSourceContexts`, `NewContent`, `DeletedContent`, `RemainingSourcesContent` — and hands the map straight to `generateWithTemplate`. That function parses the template, executes it with the full data and sends the result (`internal/application/service/wiki_ingest.go:2438`). **Nothing anywhere on this path bounds the combined size.**

For most pages that is fine. For a hub page — an entity or topic that many documents touch — it produces a failure that is not just an error but a ratchet:

1. The page grows every batch, so `ExistingContent` grows monotonically.
2. Once the assembled prompt exceeds the model context window, the provider rejects it. Providers typically surface this as **HTTP 400**.
3. `isTransientLLMError` (`wiki_ingest.go:2626`) matches `status 408`, `429`, `5xx`, `520`–`524`. **400 is not in the set**, so the failure is classified permanent and is not retried.
4. The page content is unchanged by the failed round — so it is still just as large next time. Every subsequent batch reassembles an even bigger prompt and fails the same way.

The page can therefore never converge again, and there is no mechanism that shrinks it. Smaller pages in the same knowledge base keep working, so the symptom is "wiki convergence mysteriously stalls on exactly the most important pages" rather than an obvious outage.

Related: because the failure lands only on the Wiki span, the documents `parse_status` can still read `completed` — that visibility gap is #2688. This issue is about the cause rather than the reporting.

### Observed

On a knowledge base of roughly 1,500 documents served by a self-hosted model with a 32K context window, five hub pages grew to 15K–26K characters and then stopped converging entirely, while the rest of the knowledge base continued normally. Wiki throughput on that KB collapsed to roughly 80 pages/hour, with those pages repeatedly re-entering batches, re-taking their slug locks, and failing on context length.

### Proposed Solution

Trim the reduce inputs before the call, shedding the least authoritative and most recoverable material first. A working ordering, in decreasing willingness to drop:

1. `SharedSourceContexts` → titles only
2. `RemainingSourcesContent` → trimmed
3. `NewContent` → keep each new documents required Name/Description line, trim the verbatim evidence under it
4. `ExistingContent` → trim the **middle**, keeping the head (through the first couple of sections) and the tail, with an explicit marker in place of the omitted span so the model is told not to delete sections it cannot see
5. `DeletedContent` → never trimmed; retractions must still be able to identify what to remove

I have this running in a fork and it resolved the stall. **But I do not think the budget value itself should be ported as-is, and that is what I would like guidance on.**

### The open question: where does the budget come from?

My fork hardcodes a character budget tuned to its own 32K-context model. That is wrong for upstream: a deployment on a 128K or 200K model would get its inputs trimmed for no reason, silently lowering wiki quality to fix a problem it does not have. Three options, none obviously best:

**A. Derive it from the models context window.** Correct in principle and needs no operator action. But `chat.Options.MaxTokens` is the *completion* cap, and I could not find anywhere the context window is exposed to this call site — so this needs a new capability on the chat model abstraction. Biggest change, best result.

**B. Opt-in `WikiConfig` field, `0` = disabled.** Byte-identical behavior for everyone who does not set it, and operators who hit the wall can turn it on. Small and safe, but it fixes nobody by default, and whoever hits this has to first understand the failure well enough to know the knob exists. Note this would land on the same `WikiConfig` surface as the caps proposed in #2489.

**C. A generous hardcoded default** (say a budget only pathological pages reach), no config. Helps everyone out of the box, no new surface — but it is still a guess, and it silently changes behavior for large-context deployments.

I lean **B for now, A eventually** — B is a safe stopgap that unblocks affected deployments without risking anyone else, and A can replace it once the context window is plumbed through, at which point Bs field becomes an override rather than the only source.

### Alternatives

Adding `400` to `isTransientLLMError` would be wrong — retrying an over-length prompt just burns the retry budget on a request that cannot succeed.

Capping page size (#2489) reduces how often this is reached but does not bound the *other four* inputs, so it is complementary rather than a substitute.

### Impact

Blocks a use case

### Additional Information

Happy to open the PR once maintainers pick a direction — the trimming logic and its tests are already written and I would just need to rework how the budget is resolved. If you would rather see the code first to judge, say so and I will put up B as a draft.

### Confirmation

Hughhhhcoder:
## Description
Bound the aggregate trim-eligible inputs sent to WikiPageModify with an explicit WikiConfig byte budget. This prevents hub-page reduce prompts from growing without limit while preserving deletion instructions and keeping stable source/page context where possible.

## Type of Change
- [x] 🐛 Bug fix
- [x] 🧪 Test
- [x] 📚 Documentation update

## Related Issue
Fixes #2718

## Testing
- `go test ./internal/application/service -run TestTrimWikiReduceInputsBoundsConfiguredContext -count=1`
- `go test ./internal/application/service -run "TestTrimWikiReduceInputsBoundsConfiguredContext|TestGenerateWithTemplateSetsMaxTokens" -count=1`
- `go test ./internal/application/service -count=1`
- `go test ./internal/types ./internal/application/service -count=1`
- `git diff --check`

`go test ./...` was also run; it reached the repository test execution but stopped because the host filesystem ran out of space (`no space left on device`), with no assertion failure reported. The focused and changed-package tests pass.

## Checklist
- [x] Targeted tests pass
- [ ] Full-repository check: documented above as environment-blocked
- [x] Self-reviewed the code
- [x] Added regression coverage
- [x] Updated generated documentation

### DIFF
--- a/docs/docs.go
+++ b/docs/docs.go
@@ -20734,6 +20734,10 @@ const docTemplate = `{
                     "description": "IngestReduceParallel sets the errgroup limit for the Reduce phase\n(per-slug page write) WITHIN one batch. 0 falls back to 10. Bound by the\nsame LLM concurrency / HTTP pool considerations as the Map phase, plus\nDB connection pool size. Same multiplier caveat as IngestMapParallel.",
                     "type": "integer"
                 },
+                "max_reduce_input_bytes": {
+                    "description": "MaxReduceInputBytes bounds the aggregate, trim-eligible input passed to the WikiPageModify prompt. It is an operator-provided byte budget; 0 disables the bound.",
+                    "type": "integer"
+                },
                 "max_pages_per_ingest": {
                     "description": "MaxPagesPerIngest limits pages created/updated per ingest operation (0 = no limit)",
                     "type": "integer"
--- a/internal/application/service/wiki_ingest.go
+++ b/internal/application/service/wiki_ingest.go
@@ -1289,6 +1289,10 @@ type WikiBatchContext struct {
 	ContentInstructions    string
 	ExtractionInstructions string
 
+	// ReduceInputBudgetBytes is the explicit operator-provided budget for the
+	// trim-eligible WikiPageModify prompt inputs. Zero keeps legacy behavior.
+	ReduceInputBudgetBytes int
+
 	// PlannedFolderID holds the per-slug wiki_folders.id assigned by the batch
 	// taxonomy planning pass (planBatchTaxonomy + folder resolution), keyed by
 	// page slug. Reduce applies it only to pages that aren't already filed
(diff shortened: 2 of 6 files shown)


Hughhhhcoder:
The latest `App` and `Go Lint` pull-request workflow runs are completed with `action_required`. Could a maintainer approve the workflows when convenient so CI can run? The CodeCC scan is passing and the local focused validation is complete.

lyingbug:
Thanks for looking into reduce-input growth. Trimming the middle of `ExistingContent` and then asking the model to rewrite the whole page is exactly how pages lose content (#3468). #3492 goes the other way and rejects truncated rewrites. The PR also edits the generated swagger (`docs/docs.go`), which is stale. Closing. A bounded-input approach that never drops existing page content would be welcome.

Issue #1801, opened 2026-06-25
[Bug] LLM stream request has hardcoded timeout causing response truncation on slow models


liangliwu:
## Problem
LLM streaming requests have a hardcoded HTTP client context deadline of approximately 300 seconds. When the upstream LLM generates responses slowly — e.g., taking ~12 seconds per token — the stream is cut off mid-response, returning a truncated answer with a "context deadline exceeded" error.

## Impact
- Long or complex questions receive incomplete answers
- Users see a cut-off response with no indication that it was truncated
- The SSE stream is closed abruptly (unexpected end of JSON input)
- Cannot increase `max_completion_tokens` because the timeout fires before generation completes

## Evidence
- Observed in production: Smart Reasoning query at 09:06:43→09:09:36 (173s runtime)
- Model generating at ~12s per token via vLLM backend
- Generated: 1406 chunks, only 2375 chars content, 847 thinking chunks, 0 tool_calls
- Key log lines:
  - `Failed to parse stream response: unexpected end of JSON input`
  - `Stream read error: context deadline exceeded`
- Nginx reverse proxy timeout is set to 3600s (1 hour) — not the bottleneck
- No LLM timeout configurable via environment variables (only `WEKNORA_EMBED_TIMEOUT` exists, for embedding only)

## Root Cause
The Go HTTP client context in `openai_stream.go` uses a hardcoded context deadline (around 300s by default). When the LLM generation exceeds this, the stream is killed regardless of `max_completion_tokens`.

## Related
- Issue #1738 "completions using stream=True will be limited to 30 seconds" — covers a different root cause (**client-side** Go SDK 30s default timeout + bufio token limit). This issue is about the **server-side** LLM call timeout inside WeKnora's `openai_stream.go`.

## Suggested Fix
1. Add an environment variable (e.g., `WEKNORA_LLM_STREAM_TIMEOUT`) to control the LLM stream HTTP client timeout
2. Default to 600s (10 minutes) to accommodate slower models
3. When a timeout occurs, include a descriptive error in the SSE response rather than crashing mid-stream
4. Log a warning when generation approaches the timeout threshold so operators can tune the value

lyingbug:
Reviewed current `main` (`e07e782b`): the server-side streaming timeout is now configurable, so the hardcoded-timeout defect described here can be closed as completed.

[...]

PR #2808, opened 2026-08-26
fix(agent): clamp LLMCallTimeout into a sane 10s-3600s range

oscarlius:
## Description

`AgentEngine.getLLMCallTimeout()` (`internal/agent/const.go`) previously passed any positive `AgentConfig.LLMCallTimeout` straight through to `context.WithTimeout` in `internal/agent/think.go`. Misconfigured values caused two failure modes:

- **Too small** (e.g. `1`): every LLM call times out after ~1s, before network + first-token latency can possibly complete, so the agent can never produce an answer.
- **Too large** (e.g. `86400`): the per-call guard that prevents a hung provider connection from pinning a worker for the whole pipeline is silently disabled.

This PR clamps the effective per-call timeout into `[10s, 3600s]` at the single consumption point (`getLLMCallTimeout`), so every configuration source — the agent API, custom-agent YAML, the global `WEKNORA_AGENT_LLM_TIMEOUT` env override, and the frontend form — inherits the protection regardless of where the value was written. The max matches the frontend editor's existing upper bound (`AgentEditorModal.vue`, `:max="3600"`); values `<= 0` keep the existing "use default (120s)" semantics. The comparison runs in the seconds domain to avoid `time.Duration` overflow on absurd inputs, and a nil-config guard is added as drive-by defense.

Also fixes a stale doc comment in `internal/models/chat/transport.go`: the chat/stream fallback defaults are 300s/600s (as implemented), not 600s/1800s (as documented).

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue

- Refs #916 — introduced the customizable per-agent LLM call timeout this PR hardens.
- Related: #1801 — that issue asked for a configurable LLM stream timeout (implemented in main as `WEKNORA_LLM_STREAM_TIMEOUT_SECONDS`, default 600s); the doc-comment fix in this PR corrects the documented defaults for exactly those env vars. This PR does not implement #1801's remaining asks (descriptive SSE error on timeout, near-threshold warning logs).

## Testing

Added `TestGetLLMCallTimeout` in `internal/agent/const_test.go` — table-driven, 8 cases: nil config, zero, negative, below-minimum clamp, at-minimum, in-range pass-through, at-maximum, above-maximum clamp.

- [x] `git diff --check upstream/main...HEAD` passes
- No local Go toolchain was available on the authoring machine (`go: command not found`), so the following are deferred to CI on this PR:
  - `go test ./internal/agent/ -run TestGetLLMCallTimeout -count=1`
  - `golangci-lint run --new-from-rev=origin/main ./...`
  - `gofmt` verification (edits follow existing tab-indented style)

## Checklist

- [x] Changed source files are formatted (verified visually; CI gate remains)
- [ ] Targeted tests for the changed packages/components pass — deferred to CI (see Testing)
- [ ] Diff-scoped lint passes where applicable (for Go: `golangci-lint run --new-from-rev=origin/main ./...`) — deferred to CI (see Testing)
- [x] Self-reviewed the code
- [x] Updated related documentation (inline doc comments corrected)
- [ ] Breaking changes are clearly called out in the description above — N/A, no breaking changes

### DIFF
--- a/internal/agent/const.go
+++ b/internal/agent/const.go
@@ -21,6 +21,15 @@ const (
 	// Can be overridden via AgentConfig.LLMCallTimeout.
 	defaultLLMCallTimeout = 120 * time.Second
 
+	// minLLMCallTimeout / maxLLMCallTimeout bound the effective per-call timeout
+	// when AgentConfig.LLMCallTimeout is set explicitly. Values below min can never
+	// survive network + first-token latency and almost always indicate operator
+	// error; values above max defeat the purpose of the per-call guard (a hung
+	// provider connection would pin a worker for hours). The max matches the
+	// frontend editor's upper bound (AgentEditorModal.vue, :max="3600").
+	minLLMCallTimeout = 10 * time.Second
+	maxLLMCallTimeout = 3600 * time.Second
+
 	// defaultToolExecTimeout is the default maximum time for a single tool execution.
 	// Prevents long-running tools (web_fetch, database_query) from hanging indefinitely.
 	defaultToolExecTimeout = 60 * time.Second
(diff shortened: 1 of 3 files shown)


lyingbug:
Thanks! This targets code that's since changed: `LLMCallTimeout` is now a stall timeout (`getLLMStallTimeout` in `internal/agent/const.go`, i.e. how long the stream may stay silent, not a wall-clock cap), and `internal/models/chat/transport.go` no longer exists after the model-layer refactor. The clamp rationale (a huge value pinning a worker for hours) doesn't carry over to stall semantics. Closing.

PR #2846, opened 2026-08-27
fix(sandbox): accept npipe docker hosts so Windows can use the Docker backend

qbsoft:
## Problem

On Windows the Docker sandbox backend cannot be configured at all. Leaving the daemon address blank makes `DetectLocalDockerHost` read the docker CLI context, which returns `npipe:////./pipe/docker_engine` — and `ValidateDockerHost` only knows `unix` / `tcp` / `http` / `https`, so the config is refused before it can be saved:

```
sandbox: unsupported docker host scheme "npipe"
```

The UI just shows a failed check, with no hint that it is a platform issue rather than a typo.

## Fix

Validation is the only thing in the way: the moby client this package builds already dials npipe (`client.go` handles the scheme), the outbound dial guard only concerns network endpoints, and the TLS check only applies to `tcp`/`http`/`https`. A named pipe is local to the machine, the same as a unix socket, so neither guard should touch it. Allowing the scheme is the whole fix — no transport code needed.

Also mentions `npipe://` in the missing-scheme error, which previously suggested only `unix://` or `tcp://` and left Windows users with no correct option.

## Verification

- `TestValidateDockerHost` / `TestValidateDockerRemoteTLS` cover the new scheme (valid pipe path accepted, bare `npipe://` rejected, no TLS requirement).
- Negative-verified: removing the npipe case turns `TestValidateDockerHost` red.
- Verified end-to-end on Windows 11 + Docker Desktop: with this change the Docker backend detects the local daemon, saves, and runs sandboxes.

### DIFF
--- a/internal/sandbox/docker_engine.go
+++ b/internal/sandbox/docker_engine.go
@@ -264,12 +264,28 @@ func ValidateDockerHost(host string, allowPrivate bool) error {
 	scheme, address, found := strings.Cut(trimmed, "://")
 	if !found {
 		return fmt.Errorf(
-			"sandbox: docker host %q must include a scheme (unix:// or tcp://)", host)
+			"sandbox: docker host %q must include a scheme (unix://, npipe:// or tcp://)", host)
 	}
 	switch strings.ToLower(scheme) {
 	case "unix":
 		if !strings.HasPrefix(address, "/") {
 			return fmt.Errorf("sandbox: docker unix socket path %q must be absolute", address)
+		}
+		return nil
+	case "npipe":
+		// Windows' local daemon endpoint, the named-pipe counterpart of the
+		// unix socket: npipe:////./pipe/docker_engine. It is local to this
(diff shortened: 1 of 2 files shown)


lyingbug:
Closing this implementation because the new npipe branch bypasses the private-endpoint guard for remote named pipes. On the reviewed head, ValidateDockerHost("npipe:////10.0.0.5/pipe/docker_engine", false) returns nil; I reproduced this with an additional regression probe. The Windows Docker transport passes the pipe address to go-winio, and Windows pipe names can address a remote server (https://learn.microsoft.com/en-us/windows/win32/ipc/pipe-names). Therefore, accepting every nonempty npipe address does not establish that the endpoint is local.

Windows Docker support is valuable. A replacement should validate the canonical local pipe authority/path (including rejecting remote UNC authorities) and include negative tests. Existing positive validation tests pass, but do not cover this case. This closure is based on the demonstrated validation defect, not merge conflicts; actual Windows network dialing was not exercised.

PR #2888, opened 2026-08-31
fix(skills): resolve sibling-package imports through ancestor script …

jettjia:
…dirs

The python verifier judges each file with only its own directory on sys.path, which is right for entry scripts but wrong for modules that are only ever imported by one: a real install of the xlsx skill was rejected because scripts/ecmadoc/auditors/*.py import the partkit package that sits beside auditors/ under scripts/ecmadoc/, the directory the entry script doc_audit.py puts on sys.path when it runs. The import graph inherits that path entry at runtime, so the install was sound and the check a false positive.

Give every ancestor of the checked file up to the skill root the same chance to resolve a name that the file's own directory gets. The walk only follows the file's lineage, so a module vendored under a child of an ancestor (scripts/vendor/requests.py) is still reported missing.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue

Fixes #

## Checklist
- [ ] `git diff --check origin/main...HEAD` passes
- [ ] Changed source files are formatted
- [ ] Targeted tests for the changed packages/components pass
- [ ] Self-reviewed the code

### DIFF
--- a/internal/application/service/tenant_skill_verify.py
+++ b/internal/application/service/tenant_skill_verify.py
@@ -85,11 +85,27 @@ def is_importable(name, script_dir):
     is reported without the side effects of loading a present one. script_dir
     leads the path because that is where the interpreter puts the directory of
     the script it was handed, which is how sibling modules resolve at runtime.
+
+    A file is rarely the entry point, though. Python puts the executed
+    script's directory on sys.path, and every module reached through the
+    import graph from there resolves against that same entry — so a module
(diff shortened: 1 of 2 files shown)


lyingbug:
在 #2889 中做了更详细的处理，此次PR的内容已涵盖

PR #2967, opened 2026-09-02
fix(docreader): warm the OpenDataLoader JVM at service startup

### LINKED ISSUE #1802: [Bug] opendataloader PDF parser JVM cold start causes first knowledge base import timeout
## Problem
The `opendataloader_parser.py` uses Apache Tika (via `tikajava`) for PDF parsing, which starts a JVM instance on first parse request. JVM cold start takes approximately 25-30 seconds, causing the first knowledge base document import to time out and fail.

## Impact
- First knowledge base import after service startup always fails due to JVM initialization timeout
- Users must retry the import, which then succeeds on the second attempt (JVM already running)
- Confusing user experience — first import appears broken
- The timeout is in the Go-side HTTP client calling the Python parser, which has a default timeout shorter than JVM cold start

## Evidence
- `opendataloader_parser.py` calls `tikajava.Tika()` which initializes a JVM on first use
- JVM startup time: ~25-30 seconds measured
- The Python parser is imported lazily (first request triggers import + JVM init)
- A local patch was required to: (a) eagerly import the parser module at startup, (b) use `threading.Thread` to pre-heat the JVM before the first real request arrives
- Without the patch, every container restart causes first-import failure for PDF documents

## Suggested Fix
1. In the Python parser entry point, pre-initialize the JVM at module load time (eager init) rather than on first parse call
2. Add an async pre-heat mechanism: spawn a background thread to touch the JVM on startup so it is warm before any request
3. Optionally: increase the Go-side HTTP client timeout for the Python parser service, or add a retry for JVM-cold-start failures
4. Consider adding a health-check endpoint to the parser that validates JVM readiness before the main app starts sending requests
5. Better long-term fix: use a persistent JVM pool or switch to a pure-Python PDF parser (e.g., `pypdf`) for simpler documents

LuckVd:
## Description

Fixes #1802

The docreader `opendataloader` PDF parser pays full JVM startup plus library initialization on the first `opendataloader_pdf.convert()` of each process — 25-30s on a cold container (`docreader/parser/opendataloader_parser.py:276`, `import opendataloader_pdf` happens inside `_run_convert`, and the JVM is initialized on first use). The first knowledge-base import after startup therefore exceeds the parse timeout and fails; users must retry.

This PR moves that one-time cost off the request path: right after the gRPC server starts listening, docreader runs one tiny spec-valid PDF through the normal convert path in a daemon thread. By the time the first real document arrives, the JVM and libraries are already initialized.

### Change

- New `warmup_engine()` in `docreader/parser/opendataloader_parser.py`: best-effort engine warmup that goes through the same `opendataloader_available(quick=True)` probe, `parser_worker_limit("opendataloader", ...)` slot, and `_run_convert()` path as a real parse, using an in-memory generated one-page PDF (`_minimal_warmup_pdf()`, xref offsets computed so the file is spec-valid). Every failure is returned as `(False, reason)` — warmup never raises into serving.
- `docreader/main.py`: `_start_opendataloader_warmup()` spawns the warmup as a daemon thread after `server.start()`, so startup is never blocked; failures only log.
- New `DOCREADER_ODL_WARMUP` config (default `true`), so deployments that don't want the boot-time conversion can set `DOCREADER_ODL_WARMUP=false`.

### Type of Change

- [x] 🐛 Bug fix

### Related Issue

- Fixes #1802

### Testing

- New tests in `docreader/tests/test_opendataloader_parser.py`:
  - `test_minimal_warmup_pdf_is_spec_shaped` / `test_minimal_warmup_pdf_parses_with_pypdf` — the warmup PDF is a real one-page PDF (parsed with pypdf)
  - `test_warmup_skips_when_engine_unavailable` — no convert call when java/package missing
  - `test_warmup_runs_tiny_convert_and_reports_elapsed` — exactly one convert against `warmup.pdf`, elapsed reported
  - `test_warmup_swallows_convert_failures` — convert exceptions surface as `(False, reason)`
- `pytest tests/test_opendataloader_parser.py` — 14 passed (9 pre-existing + 5 new)
- `pytest tests/` (full docreader suite) — 170 passed / 13 skipped / 14 failed; the same 14 failures (`test_excel_parser.py`, `test_mhtml_parser.py`) occur identically on a clean `origin/main` baseline (165 passed there) — pre-existing and unrelated to this change; this branch adds the 5 passing warmup tests on top.
- Graceful degradation verified live on a host without Java: `warmup_engine()` returns `(False, "需要 Java 11+（JRE）...")` in <0.1s instead of raising; the JVM-positive path runs in the standard docreader image where `openjdk-17-jre-headless` and `opendataloader-pdf` are installed.
- No Go code touched.

### Checklist

- [x] Targeted tests for the changed components pass
- [x] Self-reviewed the code
- [x] Breaking changes: none (new config defaults to enabled; set `DOCREADER_ODL_WARMUP=false` to opt out)

### DIFF
--- a/docreader/config.py
+++ b/docreader/config.py
@@ -61,6 +61,7 @@ class DocReaderConfig:
     odl_hybrid_mode: str
     odl_hybrid_fallback: bool
     odl_markdown_with_html: bool
+    odl_warmup: bool
     pdf_render_max_workers: int
     pdf_render_parallelism: int
     pdf_render_dpi: int
@@ -98,6 +99,7 @@ def load_config() -> DocReaderConfig:
     odl_markdown_with_html = _get_bool(
         ["DOCREADER_ODL_MARKDOWN_WITH_HTML"], False
     )
+    odl_warmup = _get_bool(["DOCREADER_ODL_WARMUP"], True)
     pdf_render_max_workers = _get_int(["DOCREADER_PDF_RENDER_MAX_WORKERS"], 1)
     # Intra-document render parallelism: how many worker processes render the
     # scanned pages of a SINGLE PDF in parallel. pdfium is not thread-safe, so
@@ -140,6 +142,7 @@ def load_config() -> DocReaderConfig:
         odl_hybrid_mode=odl_hybrid_mode,
         odl_hybrid_fallback=odl_hybrid_fallback,
         odl_markdown_with_html=odl_markdown_with_html,
+        odl_warmup=odl_warmup,
         pdf_render_max_workers=pdf_render_max_workers,
         pdf_render_parallelism=pdf_render_parallelism,
         pdf_render_dpi=pdf_render_dpi,
--- a/docreader/main.py
+++ b/docreader/main.py
@@ -2,6 +2,7 @@ import logging
 import os
 import re
 import sys
+import threading
 import traceback
 import uuid
 from concurrent import futures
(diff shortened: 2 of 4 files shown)


lyingbug:
Thanks! `OpenDataLoaderParser.convert()` spawns a fresh JVM on each call (see the module docstring), so warming one at startup doesn't speed up later requests. Closing.

PR #2999, opened 2026-09-02
fix(preview): prevent large markdown preview freezes

jacentsao:
## Summary
- render Markdown images with native lazy loading, async decoding, and low fetch priority
- allow the added image attributes through DOMPurify
- keep Markdown rendering on `marked.parse()` so marked-katex extensions remain active
- add regression coverage for image attributes and KaTeX rendering

## Validation
- `npx tsx --test src/components/document-preview.markdown.test.ts`
- `npm run type-check`

This keeps large Markdown as a continuous document while avoiding eager loading of every remote image.

### DIFF
--- a/frontend/src/components/document-preview.markdown.test.ts
+++ b/frontend/src/components/document-preview.markdown.test.ts
@@ -1,0 +1,30 @@
+import assert from 'node:assert/strict'
+import test from 'node:test'
+
+import { renderDocumentPreviewMarkdown } from '../utils/documentPreviewMarkdown.ts'
+import { applyDocumentPreviewImageAttributes } from '../utils/security.ts'
+
+test('preview Markdown keeps native lazy image attributes', () => {
+  const html = renderDocumentPreviewMarkdown('![test](https://example.com/test.png)', value => value)
+
+  assert.match(html, /loading="lazy"/)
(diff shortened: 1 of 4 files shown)


lyingbug:
Thanks for the lazy-load approach — I left a review-based follow-up in https://github.com/Tencent/WeKnora/pull/3075.

The main change is that image hrefs now follow the shared DOMPurify URI policy (`data:image/*`, relative, protocol-relative, provider schemes) instead of `isValidImageURL()` plus a scheme reject, so parser-emitted base64 images are not dropped. Lazy attributes stay on a preview-only sanitizer rather than the global `sanitizeHTML` allowlist.

jacentsao:
> Thanks for the lazy-load approach — I left a review-based follow-up in #3075.
> 
> The main change is that image hrefs now follow the shared DOMPurify URI policy (`data:image/*`, relative, protocol-relative, provider schemes) instead of `isValidImageURL()` plus a scheme reject, so parser-emitted base64 images are not dropped. Lazy attributes stay on a preview-only sanitizer rather than the global `sanitizeHTML` allowlist.

Thanks for the refinements, the new approach for Data URIs looks great. I'll close this one in favor of #3075.

PR #3028, opened 2026-09-04
fix(frontend): recover the chat stream from an expired access token

hjxccc:
## Problem

When the access token expires while a chat page is open, sending a message fails silently:

- the user's bubble is appended to the conversation, but no answer ever arrives;
- the only feedback is a transient `流式连接失败: Error: HTTP 401` / `Stream connection failed: Error: HTTP 401` toast;
- the app does **not** refresh the token and does **not** redirect to `/login`.

Users read this as "the app ignored me" and re-send, which stacks up duplicate bubbles that
aren't persisted server-side (they vanish on reload). The re-login prompt only appears once some
*unrelated* polling request happens to hit 401 and trips the axios interceptor — in one case we
measured that arriving more than a minute after the failed send.

## Reproduction

1. Log in and open a conversation.
2. Invalidate the access token without touching this tab — the easiest way is to sign out of the
   same account elsewhere (`Logout` calls `RevokeTokensByUserID`, which revokes *every* token
   issued to that user, so any other device signing out is enough). Letting the token expire
   naturally works too.
3. Return to the tab and send a message.

Observed: an orphaned user bubble plus an `HTTP 401` toast; no re-login prompt.
Expected: the token is refreshed and the message goes through, or the user is told to log in again.

## Root cause

`frontend/src/api/chat/streame.ts` streams over `@microsoft/fetch-event-source`, i.e. raw `fetch`.
It never passes through the axios instance in `frontend/src/utils/request.ts`, so the 401 handling
that already exists there — refresh, queue concurrent retries, clear credentials and redirect on
failure — simply does not apply to the one request that matters most in the product:

```ts
onopen: async (res) => { if (!res.ok) throw new Error(`HTTP ${res.status}`); }
onerror: (err) => { throw new Error(`${t('error.streamFailed')}: ${err}`); }
// → catch → error.value = err.message → toast, and nothing else
```

## Fix

**`utils/request.ts`** — extract the refresh out of the interceptor into an exported
`refreshAccessTokenShared()`, preserving the existing `isRefreshing` / `failedQueue` de-duplication,
and have the interceptor call it. Behaviour is unchanged; there is now one implementation instead of
one-per-transport. Also export `forceReloginRedirect()` for callers that authenticate outside axios.

**`api/chat/streame.ts`** — mark a 401 handshake with a `StreamAuthError`, and on that error refresh
once and replay the request with the new token. If the refresh fails, or a freshly minted token is
rejected again (the session was revoked, not merely expired), clear credentials, redirect to
`/login`, and surface `error.pleaseRelogin` instead of a raw status code.

## Why the replay is safe

- **It can't duplicate a message.** A 401 is produced by the auth middleware before the handler
  runs, so nothing is persisted — verified against server logs and the `messages` table: the
  rejected request left no row.
- **It's invisible.** The handshake failed, so not a single chunk has been rendered yet.
- **It happens at most once**, and only for 401. 403 and every other status keep their existing path.
- **Cancellation is respected**: the stream generation counter is re-checked after the refresh, so a
  newer send supersedes the replay.
- **Embed visitors are excluded** — they authenticate with an Embed token the SPA cannot refresh, and
  bouncing an anonymous visitor to `/login` would be wrong.
- **One refresh, not N.** Both transports now queue behind the same in-flight refresh, which matters
  because refresh tokens rotate — concurrent refreshes invalidate each other.

Control flow through the library was checked against
`@microsoft/fetch-event-source/lib/cjs/fetch.js`: `await onopen(response)` runs inside the `try`, so
a throw reaches `onerror`, and a throw from `onerror` rejects the promise. `dispose()` aborts only
the library's internal controller, leaving the caller-supplied `signal` usable for the replay.

## Testing

- `npx vue-tsc --build --force` — passes, no new errors.
- `npm test` — 577/580. The 3 failures are in `SandboxConfigEditorDrawer.network.test.mjs` and are
  pre-existing: reverting both files in this PR to their `main` content reproduces the same 3
  failures, so they are unrelated to this change.
- The 401 path itself was diagnosed from a production incident (server logs cross-checked against
  the `messages` table) rather than reproduced in a local browser session, so a maintainer sanity
  check of the refresh-and-replay flow in a running UI would be welcome.

### DIFF
--- a/frontend/src/api/chat/streame.ts
+++ b/frontend/src/api/chat/streame.ts
@@ -7,8 +7,22 @@ import {
   sanitizeStreamRequestBody,
   type StreamRequestMeta,
 } from '@/utils/chatRequestDebug';
+import { refreshAccessTokenShared, forceReloginRedirect } from '@/utils/request';
 
-
+/**
+ * The SSE handshake was rejected for auth reasons.
+ *
+ * Streaming runs on raw fetch, outside the axios instance, so nothing else
+ * refreshes the token for it. This marker lets startStream tell "your login
+ * expired" apart from a genuine stream failure and recover instead of showing
+ * a dead-end "HTTP 401" toast.
+ */
+class StreamAuthError extends Error {
+  constructor(readonly status: number) {
+    super(`HTTP ${status}`);
+    this.name = 'StreamAuthError';
+  }
+}
 
 interface StreamOptions {
   // 请求方法 (默认POST)
@@ -148,11 +162,14 @@ export function useStream() {
         sentAt: Date.now(),
       };
       
-      await fetchEventSource(url, {
+      // Wrapped so an expired access token can be refreshed and the request
+      // replayed once. Nothing has been streamed to the UI yet when the
+      // handshake 401s, so the replay is invisible to the user.
+      const runStream = (authToken: string) => fetchEventSource(url, {
         method: params.method,
         headers: {
           "Content-Type": "application/json",
-          "Authorization": embedToken ? `Embed ${embedToken}` : `Bearer ${token}`,
+          "Authorization": embedToken ? `Embed ${embedToken}` : `Bearer ${authToken}`,
           "Accept-Language": i18n.global.locale?.value || localStorage.getItem('locale') || 'zh-CN',
           "X-Request-ID": requestID,
           ...(!embedToken && tenantIdHeader ? { "X-Tenant-ID": tenantIdHeader } : {}),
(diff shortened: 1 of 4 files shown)


lyingbug:
Review follow-up pushed onto this branch:

- A second SSE handshake 401 after a successful refresh no longer wipes the newly minted session (`forceReloginRedirect`). Axios already kept the tokens in that case; workspace lookup can also 401 with a valid JWT.
- Replay is pinned to the original `AbortController`, and `StreamAuthError` is recognized by `name` as well as `instanceof`.
- Refresh/queue/retry logic now lives in `frontend/src/utils/authRefresh.ts` with unit tests.

The original expired-token refresh-and-replay path is unchanged.

PR #3029, opened 2026-09-04
feat(embedding): add persistent embedding cache with hit statistics

Li-QingS:
## Motivation

Embedding requests currently always reach the provider, even when the exact same text was embedded before. Document ingestion and RAG evaluation flows embed large parts of the corpus repeatedly (re-indexing, evaluation re-runs, repeated queries), which wastes provider calls, tokens and latency.

## What this PR adds

- A process-wide embedding cache backed by a new `embedding_cache_entries` table, keyed by `(tenant_id, model_id, dimension, SHA-256(text))` — entries are never shared across tenants or models.
- A `cachedEmbedder` wrapper covering single, batch, and batch-with-pool embedding paths; only cache misses are sent to the provider, and hits bump the entry's hit counter.
- Per-model hit / miss / provider-call statistics exposed via `GET /api/v1/embedding-cache/stats` (admin only), so operators can verify reuse and estimate savings.
- Opt-in switch: off by default, enable with `EMBEDDING_CACHE_ENABLED=true` (documented in `.env.example`).
- Migrations: `migrations/versioned/000091` (PostgreSQL) and `migrations/sqlite/000013` (Lite mode).

## Testing

- `go build ./...`
- `go test ./internal/models/embedding/... ./internal/application/repository/... ./internal/handler/... ./internal/database/...`
- End-to-end cold/hot verification with a fixed corpus: embedding the same corpus twice keeps provider calls flat on the second run (19 provider calls in both runs; 0 hits / 31 misses cold, 31 hits / 0 misses hot).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -756,3 +756,6 @@ LANGFUSE_HOST=http://langfuse-web:3000
 # WEKNORA_TRUSTED_PROXIES=
 # ants 协程池大小（Embedding 并发，出现 429 时调小）。
 CONCURRENCY_POOL_SIZE=5
+
+# Embedding cache switch. Default off; set to true to enable.
+# EMBEDDING_CACHE_ENABLED=false
--- a/internal/application/repository/embedding_cache.go
+++ b/internal/application/repository/embedding_cache.go
@@ -1,0 +1,100 @@
+package repository
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"time"
+
+	"github.com/Tencent/WeKnora/internal/types"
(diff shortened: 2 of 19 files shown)


lyingbug:
Should we use Redis with a TTL instead of the database for storage?

lyingbug:
Thanks for the work! We're consolidating the embedding caches on #3483: it's opt-in, uses Redis/memory with a TTL (the direction suggested in the review here), and keys on tenant, model config and update time. This PR also reuses migration numbers 000091/000013, which are already taken on main and would fail at startup. Closing in favor of #3483.

PR #3102, opened 2026-09-08
fix(chat): split inline <think> blocks out of the chat stream

### LINKED ISSUE #3099: [Bug]: 思考里面为啥显示了 </think>
### Affected Component

Frontend UI

### Bug Description

Image

### Steps to Reproduce

思考输出的时候

### Expected Behavior

应该优化下

### Actual Behavior

应该优化下

### WeKnora Version

v0.80

### Deployment Method

Docker Compose

### Operating System

linux

### Relevant Logs

```shell

```

### Confirmation

### LINKED ISSUE #3132: [Bug]: 输出的长文本markdown结果， 会被截断到 会话-->思考中， 造成输出不完整
### Affected Component

Frontend UI

### Bug Description

输出的长文本markdown结果， 会被截断到 会话-->思考中， 造成输出不完整

Image

### Steps to Reproduce

会话-->问题-->输出截断

### Expected Behavior

  完整的 问答结果

### Actual Behavior

  只输出 截断后新生成的文档，
截断前的文档 在思考中

### WeKnora Version

v0.8.0

### Deployment Method

Docker Compose

### Operating System

Ubuntu22.04

### Relevant Logs

```shell

```

### Confirmation

LuckVd:
## Description

Fixes #3099
Fixes #3132

The issue reports a literal `</think>` tag rendering inside the chat thinking card (v0.8.0, Docker Compose). Code-level confirmation of two contributing defects:

1. **Backend**: the agent pipeline splits inline `<think>…</think>` reasoning out of the plain content channel (`internal/agent/think.go:417-435`), but the chat pipeline forwarded content-channel chunks verbatim (`internal/application/service/chat_pipeline/chat_completion_stream.go`). Providers whose thinking toggle cannot be disabled server-side (Minimax-M2 per the existing comment in `internal/models/chat/openai_stream.go:25-26`, plus several vLLM/Ollama Qwen deployments) emit the tags in that channel, so the raw markup reached the client and the saved history.
2. **Frontend**: `frontend/src/composables/useChatStreamHandler.ts` parsed the accumulated content with one `indexOf('<think>')` / `lastIndexOf('</think>')` / single-shot `replace('<think>', '')` pass. With more than one think block (one per reasoning round) the text before the *last* close tag — including earlier blocks and their tags — became the "thinking" body, which is exactly the leaked `</think>` in the screenshot.

### Change

- Backend: answer-channel chunks now pass through `ThinkStreamSplitter` before the resource-aware answer decoder; reasoning segments are emitted as thinking events, mirroring the agent path.
- Backend: the thinking card is closed only when real answer text starts (or the stream ends), so it does not flicker while inline think and answer alternate across chunks.
- Frontend: new `parseThinkBlocks` util (`frontend/src/utils/thinkBlocks.ts`) parses any number of interleaved blocks; `useChatStreamHandler.ts` now uses it for both the streaming path and the history-normalization path (legacy sessions saved with raw tags render correctly again).
- Frontend (follow-up, #3132): `parseThinkBlocks` is now code-region aware — think-tag markup inside fenced code blocks (``` / ~~~, up to 3 leading spaces, longer closes allowed) and inline code spans (equal-length backtick runs) is treated as literal document content, so a long markdown answer that merely *mentions* the tags no longer gets its head chopped into the thinking card. Unterminated fences/spans consume the rest as literal answer, mirroring markdown rendering. Inside a real block only the close tag matters, so reasoning containing fences is unaffected.

### Type of Change

- [x] 🐛 Bug fix

### Related Issue

- Fixes #3099
- Fixes #3132

### Testing

- New tests:
  - Go: `TestStreamSplitsInlineThinkBlocks`, `TestStreamSplitsMultipleInlineThinkBlocks`, `TestStreamUnterminatedThinkBlockGoesToThought` (failure path: unterminated block at EOS is thinking text, not answer), `TestStreamPlainAnswerUnchanged` (no-tags path byte-identical to before).
  - Frontend: `frontend/src/utils/thinkBlocks.test.ts` — 19 cases covering passthrough, single/multiple blocks, streaming growth, multiline reasoning, empty blocks, stray close tags, `hasThinkMarkup`, plus #3132 cases: tags inside fenced blocks / tilde fences / longer closes / inline code spans stay literal, unterminated fence & span consume the rest, a real block before a fenced example still parses, streaming growth where a fence closes later and the real block pops out.
- `go test ./internal/application/service/chat_pipeline/ -count=1` — pass (full package, incl. the two pre-existing stream tests).
- `go vet ./internal/application/service/chat_pipeline/` — clean
- `gofmt -l` no output; `git diff --check` clean
- Frontend: `node --experimental-strip-types --test src/utils/thinkBlocks.test.ts` — 19/19 pass (repo's test runner is `tsx --test`; run with the node type-stripping flag since no node_modules are installed in this environment; the util has no dependencies beyond `node:test`)
- Pre-existing issue note: none — no failures observed on the baseline for the touched packages.
- Not covered: a live browser run of the embed/chat UI (no deployment environment here); the changed frontend logic is exercised through the unit tests above.

### Checklist

- [x] Self-reviewed the code
- [ ] Breaking changes are clearly called out in the description above — none

### DIFF
--- a/frontend/src/composables/useChatStreamHandler.ts
+++ b/frontend/src/composables/useChatStreamHandler.ts
@@ -4,6 +4,7 @@ import { useI18n } from 'vue-i18n'
 import { ensureRagPipelineHistoryStream } from '@/utils/rag-pipeline-history'
 import { applyMessageCreatedAt, bindServerTurnTimestamps, ensureMessageCreatedAt } from '@/utils/messageTimestamp'
 import { expandSteerForksInHistory, forkAfterInjectedUser, steerStepEvents, resetSteerTurnForReplay } from '@/utils/steerStreamFork'
+import { hasThinkMarkup, parseThinkBlocks } from '@/utils/thinkBlocks'
 
 export type ChatMessage = Record<string, unknown>
 
@@ -464,23 +465,19 @@ export function useChatStreamHandler(options: UseChatStreamHandlerOptions) {
       restoreQuickAnswerFlags(item)
 
       if (item.content) {
-        const content = String(item.content)
-        const thinkCloseTag = '</think>'
-        if (!content.includes('<think>') && !content.includes(thinkCloseTag)) {
+        // Models that embed <think>…</think> in the content channel may emit
+        // several blocks (one per reasoning round). Parse globally so inner
+        // tags never leak into the thinking card (#3099).
+        const parsed = parseThinkBlocks(String(item.content))
+        if (hasThinkMarkup(String(item.content))) {
+          item.thinkContent = parsed.think
+          item.showThink = Boolean(parsed.think) || parsed.thinking
+          item.thinking = parsed.thinking
+          item.content = parsed.thinking ? '' : parsed.answer
+        } else {
           item.thinkContent = ''
           item.showThink = false
           item.thinking = false
-        } else if (content.includes(thinkCloseTag)) {
-          item.showThink = true
-          item.thinking = false
-          const index = content.trim().lastIndexOf(thinkCloseTag)
-          item.thinkContent = content.trim().substring(0, index).replace('<think>', '').trim()
-          item.content = content.trim().substring(index + thinkCloseTag.length)
-        } else if (content.includes('<think>')) {
-          item.showThink = true
-          item.thinking = true
-          item.thinkContent = content.replace('<think>', '').trim()
-          item.content = ''
         }
       }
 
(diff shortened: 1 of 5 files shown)


LuckVd:
Follow-up commit `06ebb1c3` (pushed to this branch) extends `parseThinkBlocks` with code-region awareness: think-tag markup inside fenced code blocks (backtick and tilde fences, up to 3 leading spaces, longer closes allowed) and inline code spans (equal-length backtick runs) is now treated as literal document content. This also fixes #3132 — long markdown answers that merely *mention* the tags were split at the literal tag, chopping the head of the document into the thinking card. Unterminated fences/spans consume the rest as literal answer, mirroring how a markdown renderer treats an unclosed fence; inside a real block only the close tag matters, so reasoning containing fences is unaffected. Tests: 19/19 in `thinkBlocks.test.ts`.

LuckVd:
Follow-up commit a9345b98 fixes the backend half of #3132 that the earlier commits missed.

**What was wrong:** the code-region awareness added earlier only lived in the frontend parser (`parseThinkBlocks`). On the backend, `ThinkStreamSplitter` (used by both the agent pipeline and the chat pipeline added in this PR) and `StripThinkBlocks` still matched `<think>`/`</think>` as bare literals. So during **live streaming**, an answer that merely mentions the tags inside a fenced code block or inline code span still had its head routed into thinking events — the exact #3132 symptom — and the regex in `StripThinkBlocks` deleted the tagged span from the stored message. The frontend re-parse could only repair history rendering, not the live stream or the stored content. That is why the reporter still saw the problem.

**What changed:**
- `ThinkStreamSplitter` now tracks fenced code blocks (``` / ~~~, ≥3 chars, up to 3 leading spaces) and inline code spans (exact backtick runs) at the top level and treats tags inside them as literal answer text — the same rules as `frontend/src/utils/thinkBlocks.ts`, so live stream and re-parsed history agree. Chunk boundaries straddling a tag / fence line / backtick run are handled by holding back the inconclusive tail until the next Feed/Flush.
- `StripThinkBlocks` now runs through the same parser instead of a bare regex, so stored content keeps code-region mentions.

**Verification:** `go test ./internal/agent/tools/ ./internal/agent/ ./internal/application/service/chat_pipeline/` all pass; new tests cover fenced/tilde fences, inline spans, tags and fences split across chunks, unterminated fence/span, and identical routing across chunk sizes 1–11; `go vet` and `gofmt` clean.

@MosherJ the fix is on this PR branch — if you can build a frontend+backend image from this branch (or wait for the PR to be merged into main), your case should be resolved; if you still see truncation after that, please attach the raw model output (the message content in the DB) so we can see the exact split point.

PR #3104, opened 2026-09-08
fix(agent): stop emitting terminal answer events for retryable empty stops

### LINKED ISSUE #2906: [Bug]: Empty agent retries emit premature terminal answer events
### Affected Component

Backend Service & API / Agent event lifecycle

### Bug Description

When an Agent LLM call stops naturally with empty content and no tool calls, WeKnora correctly decides to retry with a nudge. However, `analyzeResponse` emits `EventAgentFinalAnswer` with `Done=true` before returning `emptyContent=true` to the retry loop.

Downstream consumers therefore observe a terminal answer while the Agent is still running. The IM consumer closes its answer-done channel on that event and waits only a bounded time for `EventAgentComplete`; if the retry is slower than that window, IM can finalize a fallback response and cancel the still-running Agent.

This is still present at upstream commit `af0aeb7cc284d76e2c59b4eaf686132e0d5b8caa`: `internal/agent/observe.go` emits `Done=true` before returning `emptyContent: response.Content == ""`.

This is related to #818 / #819, which added empty-response retries, and to the terminal-event semantics discussed in #2339, but it is a distinct downstream race.

### Steps to Reproduce

1. Run an Agent-mode QA request through an event consumer such as IM.
2. Make the first model response contain no content and no tool calls with a natural stop finish reason.
3. Make the subsequent retry take longer than the consumer's completion wait window.
4. Observe `EventAgentFinalAnswer{Content:"", Done:true}` before the retry completes.

A deterministic unit reproduction can provide an empty first response followed by a non-empty response and count terminal `EventAgentFinalAnswer` events.

### Expected Behavior

Retryable empty responses should emit no terminal answer event. Exactly one terminal answer event should be emitted after a non-empty retry succeeds or after retries are exhausted and a fallback answer is selected.

### Actual Behavior

Each empty natural stop emits `Done=true` before the retry decision, allowing downstream consumers to terminate early.

### WeKnora Version

- `v0.7.2`
- Confirmed in upstream `main` at `af0aeb7cc284d76e2c59b4eaf686132e0d5b8caa`

### Deployment Method

Built from source; reproducible in Agent unit tests and IM event consumption

### Operating System

Platform-independent backend behavior

### Relevant Logs

No specific error log is required. The failure is an ordering problem between the final-answer event and the empty-response retry decision.

### Proposed Fix

Detect an empty natural stop before emitting final-answer content or `Done`. If retries are exhausted, emit the selected fallback as the sole terminal answer. Add regression assertions that retry attempts produce exactly one terminal event in total.

### Confirmation

- [x] I searched existing issues and confirmed this exact retry/terminal ordering bug is not already reported.

vicfei:
Fixes #2906

## Problem

As reported in #2906: when the agent LLM stops naturally with empty content and no tool calls, `analyzeResponse` emits `EventAgentFinalAnswer{Done: true}` **before** returning `emptyContent: true` to the retry loop. Every retryable empty stop therefore produced a premature terminal answer event while the agent was still running. Downstream consumers that finalize on `Done=true` — e.g. the IM channel, which closes its answer-done channel and then only waits a bounded time for `EventAgentComplete` — can lock in a fallback and cancel the still-running agent when the nudged retry is slower than that window.

Verified on current `main` (`internal/agent/observe.go`, Case 1): the `Done: true` marker is emitted unconditionally, including the `response.Content == ""` path that the caller then retries.

## Change

- `observe.go`: an empty natural stop now emits **no** `EventAgentFinalAnswer` at all — the verdict's `emptyContent` alone tells the caller the round is retryable.
- `engine.go`: when empty-response retries are exhausted, the loop emits the selected fallback as the turn's **sole** terminal answer (content + `Done: true`), mirroring the non-streamed answer emission shape.

Net effect: an agent turn always ends with **exactly one** terminal answer event — the real answer after a successful retry, or the fallback after exhausted retries — matching the expected behavior in the issue.

## Tests

Three regression tests in `internal/agent/terminal_event_test.go` (all verified **red on the pre-fix code**):

1. `TestAnalyzeResponse_EmptyNaturalStop_EmitsNoTerminalAnswer` — the empty verdict emits no answer events at all.
2. `TestExecuteLoop_EmptyThenAnswer_EmitsSingleTerminalAnswer` — the deterministic repro from the issue: empty first response, real answer on retry → exactly one terminal event carrying the real answer.
3. `TestExecuteLoop_AllEmptyResponses_FallbackIsSoleTerminalAnswer` — exhausted retries → exactly one terminal event carrying the fallback.

Full `internal/agent` suite passes; repo-wide `go build` clean. (Two pre-existing Python-environment failures in skill-runtime tests reproduce on pristine `main` on this host and are unrelated.)

### DIFF
--- a/internal/agent/engine.go
+++ b/internal/agent/engine.go
@@ -726,10 +726,33 @@ func (e *AgentEngine) runReActIteration(
 				})
 				return iterOutcomeContinue, nil
 			}
-			// Retries exhausted — use fallback message rather than empty answer
+			// Retries exhausted — use fallback message rather than empty answer.
+			// analyzeResponse emitted nothing for the empty rounds (they were
+			// retryable), so the fallback must be emitted here as the turn's
+			// sole terminal answer event (#2906).
 			logger.Warnf(ctx, "[Agent][Round-%d] Empty content after %d retries - using fallback",
 				round, maxEmptyResponseRetries)
-			state.FinalAnswer = "I'm sorry, I was unable to generate a response. Please try again."
+			fallback := "I'm sorry, I was unable to generate a response. Please try again."
+			answerID := generateEventID("answer")
+			_ = e.eventBus.Emit(ctx, event.Event{
+				ID:        answerID,
+				Type:      event.EventAgentFinalAnswer,
+				SessionID: sessionID,
+				Data: event.AgentFinalAnswerData{
+					Content: fallback,
+					Done:    false,
+				},
+			})
+			_ = e.eventBus.Emit(ctx, event.Event{
+				ID:        answerID,
+				Type:      event.EventAgentFinalAnswer,
+				SessionID: sessionID,
+				Data: event.AgentFinalAnswerData{
+					Content: "",
+					Done:    true,
+				},
+			})
+			state.FinalAnswer = fallback
 			state.IsComplete = true
 			state.RoundSteps = append(state.RoundSteps, verdict.step)
 			return iterOutcomeBreak, nil
--- a/internal/agent/observe.go
+++ b/internal/agent/observe.go
@@ -408,6 +408,21 @@ func (e *AgentEngine) analyzeResponse(
 			"answer_len": len(response.Content),
 		})
 
+		// An empty natural stop is retryable (the caller nudges the model and
+		// runs another round), so it must not emit any terminal answer event
+		// yet: downstream consumers treat a Done=true EventAgentFinalAnswer as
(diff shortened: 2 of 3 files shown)


vicfei:
Gentle nudge on this one — CI has been green for several days now. It addresses both defects from #2906 (the premature terminal answer event on empty retries, and the missing fallback terminal answer when retries are exhausted), with regression tests that were red on the unfixed code. Happy to adjust anything a review turns up.

PR #3119, opened 2026-09-08
feat(frontend): remove zh-CN and ko-KR locales from web UI

VictorMaznyuk:
## Summary
- Remove 简体中文 (zh-CN) and 한국어 (ko-KR) from the language selectors (login page, general settings, embed channel panel)
- Drop their vue-i18n message registrations, embed widget locales, and TDesign locale map entries
- Supported locale set is now en-US, ru-RU, ja-JP with en-US as default/fallback

Locale resource files for zh-CN/ko-KR remain in the tree since they are still referenced by locale key audit tooling.

## Test plan
- [x] `npm run type-check` clean
- [x] `node --test src/i18n/resolveDefaultLocale.test.ts` — 4/4 pass
- [x] `npm run build` succeeds; served bundle contains no 简体中文/한국어 selector entries

🤖 Generated with [Claude Code](https://claude.com/claude-code)


VictorMaznyuk:
Closing: moving to a standalone repository instead of contributing upstream.

PR #3125, opened 2026-09-09
perf(retrieval): BM25 result cache, embedding retry hardening, query LRU+singleflight, ivfflat probes

Smith-106:
# Summary

Perf/reliability improvements to the retrieval hot path, extracted from a production deployment at scale (thousands of KBs per tenant, high query concurrency).

## BM25/Keywords result cache (`postgres` retriever)

- Redis result cache keyed by `tenant + query + filters + topK + corpus_version`; `HINCRBY corpus_ver` on every index write keeps invalidation exact (no TTL-only staleness); TTL 600s, fail-open on Redis errors, skipped for queries returning >1000 rows.

## Embedding call hardening

- `Embed`: removed the outer `for-range-3` retry loop (worst case 12 → 4 requests); honors 429 `Retry-After` in `doRequestWithRetry`.
- `GetQueryEmbedding`: in-process LRU (2k entries, 24h TTL) + per-key singleflight lock to dedupe concurrent identical queries (same model + same query text → same vector).

## ivfflat probes

- `VectorRetrieve`: `SET LOCAL ivfflat.probes = 8` for dimension > 4000 (ivfflat partial index), keeping the `hnsw.ef_search` path untouched; fallback error matching extended to include ivfflat probe errors.

## Wiring

- `pgRepository` / engine factory: inject `*redis.Client` (nil-safe, Lite mode unchanged); new `internal/common/lru.go` and `internal/application/service/embedding_locks.go` (no external deps added).

# Test plan

- `go vet` clean; new tests: `embedding_locks_test.go`, `lru_test.go`, `redislock/failure_test.go`, `openai_retry_test.go`, `repository_failopen_test.go` (422 lines: cache fail-open / corpus-version unreadable paths).
- All touched-package tests pass locally (`service`, `common`, `redislock`, `models/embedding` OpenAI suite, `postgres` retriever); remaining local failures are Windows-only (cgo sqlite3.h / fake-ip DNS) and unrelated to this change.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,16 +1,9 @@
 services:
   frontend:
     image: wechatopenai/weknora-ui:${WEKNORA_VERSION:-latest}
-    # Multi-stage Dockerfile builds dist inside the image (no host npm required).
-    # Optional: NPM_REGISTRY=https://registry.npmmirror.com docker compose build frontend
-    # VITE_FRONTEND_COMMIT: start_all.sh / make docker-build-frontend inject git;
-    # otherwise pass it explicitly or the system-info page shows "unknown".
+    # Requires frontend/dist — run ./scripts/build_frontend_dist.sh before --build
     build:
       context: ./frontend
-      args:
-        VITE_FRONTEND_COMMIT: ${VITE_FRONTEND_COMMIT:-unknown}
-        NPM_REGISTRY: ${NPM_REGISTRY:-}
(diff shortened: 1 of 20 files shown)


lyingbug:
Thanks for the work! This bundles four unrelated changes (BM25 result cache, embedding retry hardening, query LRU + singleflight, ivfflat probes), which makes it hard to review and land. It also references an `embedding_3584` column that no migration creates, and the embedding part overlaps with the model-layer refactor now in progress. Closing. Please split it into focused PRs, starting with whichever gives the biggest measurable win, with numbers.

Issue #3093, opened 2026-09-07
Ranking core (compositeScore / applyMMR) is implemented twice — chat pipeline vs agent knowledge_search — and the two copies have diverged


vicfei:
### Affected Component

Backend — retrieval ranking shared by both Q&A surfaces:

- `internal/application/service/chat_pipeline/rerank.go` (chat / knowledge-chat pipeline)
- `internal/agent/tools/knowledge_search.go` (ReAct agent `knowledge_search` tool)

### Problem

The ranking core — composite scoring (`0.6*modelScore + 0.3*baseScore + 0.1*sourceWeight`, clamped to [0,1]) and MMR selection with Jaccard token-set redundancy — exists as **two independent implementations**, and they have already drifted:

| | chat pipeline | agent tool |
|---|---|---|
| compositeScore | `chat_pipeline/rerank.go:439` | `knowledge_search.go:1210` |
| applyMMR | `chat_pipeline/rerank.go:458` | `knowledge_search.go:1249` |
| position prior | **not applied** | **applied** (`positionPrior` ±0.05 from `StartAt/EndAt`) |
| MMR structure | naive per-candidate redundancy scan, `ParallelMap` tokenization | incremental `maxRedundancy` cache (comment asserts identical selection output) |

The practical consequence of the drift: **the same retrieved chunk can rank differently depending on whether the answer is produced through a chat session or an agent session**, and every future ranking change (weights, normalization, source weighting) has to be made twice or silently diverge further.

### Why consolidation is straightforward

- Both copies already sit on the same shared primitives: the agent's `tokenizeSimple` (`knowledge_search.go:1330`) is a pure delegate to `searchutil.TokenizeSimple`, `clampFloat` delegates to `searchutil.ClampFloat`, and both MMRs use `searchutil.Jaccard`.
- The agent's `searchResultWithMeta` (`knowledge_search.go:110`) embeds `*types.SearchResult`, so a shared implementation can operate on the common type.

### Proposal

1. Extract `compositeScore` / `applyMMR` into the existing shared package (e.g. `searchutil` or a small ranking helper package).
2. Keep path-specific concerns (pipeline event emission vs logger, parallel tokenization) as parameters/options.
3. The one semantic decision for maintainers: should the agent's position prior be unified into both paths or stay agent-only? Either way, making the choice *once* in shared code is the point.

[...]

PR #3150, opened 2026-09-10
refactor(ranking): unify the duplicated compositeScore/applyMMR core into searchutil

vicfei:
Implements #3093.

## What changes

The ranking core that had drifted between the chat pipeline and the agent
`knowledge_search` tool now lives once in `internal/searchutil/ranking.go`:

- **`CompositeScore` / `CompositeRawScore`** — the `0.6*modelScore +
  0.3*baseScore + 0.1*sourceWeight` blend with the `web_search` source
  weight. The raw variant lets the agent multiply in its position prior
  before the final clamp, exactly as it did before.
- **`PositionPrior`** — the agent-only `StartAt/EndAt` prior. It stays
  agent-side in this PR; whether it should also apply to the chat pipeline
  is the open decision from the issue, now changeable in one place.
- **`ApplyMMR`** — generic MMR over precomputed token sets using the
  incremental `maxRedundancy` form. It returns the selection plus the
  average pairwise redundancy so each caller keeps its own logging and
  tokenization strategy (the chat pipeline tokenizes enriched passages in
  parallel and emits `mmr_start`/`mmr_done` events; the agent tool
  tokenizes sequentially and logs via its logger).

Path-specific wrappers shrink to their concerns:

- `chat_pipeline/rerank.go`: `compositeScore` is gone (one-line call at the
  rerank loop), `applyMMR` keeps events + `ParallelMap` tokenization, and
  its unused `chatManage` parameter is dropped.
- `agent/tools/knowledge_search.go`: `compositeScore`/`applyMMR` delegate to
  the shared core; the `tokenizeSimple`/`clampFloat`/`jaccard` one-line
  delegates are removed.

## No behavior change

- The agent multiplies the prior into the **raw** score and clamps once at
  the end — the same single clamp as before, including for out-of-range
  inputs.
- The incremental MMR form is pinned against the naive rescan form by:
  - a randomized differential test (300 seeded cases, exact selection
    order and average redundancy) in `searchutil/ranking_test.go`,
  - the existing corpus-based `TestApplyMMR_matchesNaiveSelection` in the
    agent tools, which now exercises the wrapper on top of the shared core.

## Left out on purpose

`grep_chunks.go` carries a third `applyMMR` variant whose swap-based
candidate removal changes tie-breaking relative to the order-preserving
forms consolidated here. Unifying it would be a (tiny) behavior change, so
it is deferred to a follow-up.

## Test plan

- [x] `go build ./...`
- [x] `go vet` on the three touched packages
- [x] `go test ./internal/searchutil/ ./internal/application/service/chat_pipeline/ ./internal/agent/tools/`
      — all green except the pre-existing, environment-dependent
      `TestSkillPythonPackageRecoveryInstallsWithoutPip/ensurepip`, which
      fails identically on pristine upstream/main on this host (missing
      python3-venv) and is unrelated to this change.

### DIFF
--- a/internal/agent/tools/knowledge_search.go
+++ b/internal/agent/tools/knowledge_search.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"math"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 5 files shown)


vicfei:
The failing check is `TestMCPToolApprovalConcurrentColumnPatches` (`internal/application/repository/mcp_tool_approval_repository_test.go:113`), which this PR does not touch (the diff is confined to `internal/searchutil`, `internal/agent/tools`, and `internal/application/service/chat_pipeline`).

It looks flaky on current main rather than related to this change:

- the same test at the same line also failed ~3.5h later on an unrelated PR's run (run 34441808207, branch `codex/feat-faq-enabled-f`, different code);
- on this branch the full `internal/application/repository` package passes locally twice in a row (`go test ./internal/application/repository/ -count=1`).

I don't have permission to re-run the workflow. Happy to rebase or push a no-op commit if that's easier for a re-run.

lyingbug:
Thanks! Deduplicating `compositeScore`/`applyMMR` is still worthwhile, but this branch is stale: #3381 renamed `knowledge_search.go` to `search_knowledge.go` and reworked the retrieval tools around it, so it would need a rewrite rather than a rebase. Closing. A fresh PR against current main would be welcome.

PR #3163, opened 2026-09-10
feat(faq): allow management search to include disabled entries

### LINKED ISSUE #3159: [Feature]: Allow FAQ management search to include disabled entries
### Affected Component

Backend Service & API

### Problem Description

The FAQ management search endpoint currently retrieves only enabled FAQ entries. This is correct for model-visible retrieval, but it prevents administrators from using semantic search to inspect disabled entries before editing or re-enabling them.

The paginated management list can now filter by enabled state through #3157, but keyword/list filtering does not provide the same semantic matching behavior as `POST /api/v1/knowledge-bases/{id}/faq/search`.

Temporarily enabling an entry just to test whether it matches a query is unsafe because the entry may become available to chat or Agent retrieval during that interval.

### Proposed Solution

Add an optional `include_disabled` boolean field to the FAQ management search request:

```json
{
  "query_text": "How do I request a refund?",
  "include_disabled": true
}
```

Suggested semantics:

- Omitted or `false`: retrieve enabled FAQ entries only, preserving the current behavior.
- `true`: retrieve both enabled and disabled FAQ entries.
- Return each entry's existing `is_enabled` field so callers can distinguish its state.

The option should apply only to `POST /api/v1/knowledge-bases/{id}/faq/search`. It must not change chat, Agent, knowledge-base QA, or any other model-visible retrieval path.

Implementation should preserve the existing vector-search behavior and apply the option consistently across supported retrieval drivers. Default retrieval must continue to enforce `is_enabled = true`, including post-retrieval hydration safeguards.

### Alternatives

1. Use the paginated FAQ entry list with `is_enabled=false`. This supports administration but does not provide vector similarity search, score, threshold, or matched-question behavior.
2. Temporarily enable disabled entries for testing. This can expose them to production retrieval and is therefore unsafe.
3. Add a tri-state `is_enabled` search field. The immediate use case only needs the backward-compatible distinction between enabled-only and include-disabled, so a separate three-state filter would add unnecessary API complexity.

### Impact

My team or a small group

### Use Case

An administrator is reviewing disabled FAQ entries and wants to test which entries would match a representative user question before deciding whether to edit, re-enable, or delete them. The search result must identify disabled entries without making them available to end-user answers.

### Additional Information

Related work:

- #1450 established that disabled FAQ entries must not be returned by end-user QA retrieval.
- #2392 enforced that rule across model-visible retrieval paths while preserving management access.
- #3154 and #3157 added enabled-state filtering to the FAQ management list.

This proposal preserves those guarantees. It adds an explicit opt-in capability only to the dedicated FAQ management search endpoint.

### Confirmation

- [x] I have searched existing issues and confirmed this is a new request.
- [x] I understand this request may need discussion and evaluation.

langcaiye:
## Description

Add an optional `include_disabled` parameter to the FAQ management search endpoint:

`POST /api/v1/knowledge-bases/{id}/faq/search`

- omitted or `false`: keep the existing enabled-only behavior
- `true`: search both enabled and disabled FAQ entries

This is intended for management workflows such as locating and re-enabling a disabled FAQ entry.

## Safety and scope

- The option is bound only from `FAQSearchRequest`.
- The internal `SearchParams.IncludeDisabled` field uses `json:"-"`, so generic search endpoints cannot enable it through request JSON.
- Chat, Agent, and knowledge-base QA retrieval continue to use the default enabled-only behavior.
- The enabled filter is conditionally omitted across the supported retrieval backends: PostgreSQL, SQLite, Qdrant, Milvus, Weaviate, Doris, Tencent VectorDB, Elasticsearch v7/v8, and OpenSearch.
- Result hydration keeps the enabled-only post-filter by default and relaxes it only for this explicit management request.

The API documentation, Swagger schema, Go client, and frontend API type are updated. No FAQ management UI behavior is changed in this PR.

## Tests

- `go test ./internal/types ./internal/handler ./internal/application/service ./internal/application/repository/retriever/postgres ./internal/application/repository/retriever/sqlite ./internal/application/repository/retriever/qdrant ./internal/application/repository/retriever/milvus ./internal/application/repository/retriever/weaviate ./internal/application/repository/retriever/doris ./internal/application/repository/retriever/tencentvectordb ./internal/application/repository/retriever/elasticsearch/v7 ./internal/application/repository/retriever/elasticsearch/v8 ./internal/application/repository/retriever/opensearch ./docs -count=1`
- `go test ./... -run '^$' -count=1`
- `go vet` on the changed Go packages
- `go test ./... -run '^$' -count=1` in `client/`
- `npm run type-check` in `frontend/`

Fixes #3159


lyingbug:
Thanks! This changes 10 retriever backends across 37 files but adds no UI, and disabled entries can already be found through management listing: `GET /faq/entries?keyword=...&is_enabled=false`. The benefit doesn't justify the footprint, and the branch conflicts. Closing.

PR #3183, opened 2026-09-11
fix(agent): honor MCP service advanced_config.timeout for CallTool window

### LINKED ISSUE #3135: [Bug]: MCP tool calls are capped by the generic 60s tool timeout, ignoring the service's configured advanced_config.timeout
### Affected Component

Backend Service & API

### Bug Description

An MCP service carries a user-configurable `advanced_config.timeout` (default `30`, see `types.GetDefaultAdvancedConfig`). That value is honored by the transport layers:

- `internal/mcp/client.go:165-166` — MCP HTTP client timeout
- `internal/mcp/manager.go:190-194` — `initialize` handshake timeout (capped at 60s)

But the agent-side per-call budget never sees it. `internal/agent/const.go:55-60` decides the budget from the **tool name alone**:

```go
func toolExecutionTimeout(toolName string) time.Duration {
	if toolName == "shell_exec" {
		return shellExecToolTimeout
	}
	return defaultToolExecTimeout // 60s
}
```

and `internal/agent/act.go:528` applies it to every tool call, MCP tools included:

```go
execTimeout := toolExecutionTimeout(tc.Function.Name)
...
execCtx, toolCancel := context.WithTimeout(toolExecCtx, execTimeout)
```

The post-approval path in `internal/agent/tools/mcp_tool.go:204-211` re-derives a fresh window from `meta.ExecTimeout`, i.e. the same 60s, with a hard-coded 60s fallback when it is unset.

So an operator can raise an MCP service to `timeout: 300`, watch the transport honor it, and still have every individual tool call cancelled at 60s. The configuration is silently partially applied.

The service object is already reachable at both places — `MCPTool` holds `service *types.MCPService` (`internal/agent/tools/mcp_tool.go:22-24`), and `act.go` already resolves MCP targets through the registry (`e.toolRegistry.MCPCallTarget(...)`, `act.go:464`). No new plumbing or configuration surface is needed.

### Steps to Reproduce

1. Stand up an MCP server exposing one tool that deliberately takes ~90s to return.
2. Register it in WeKnora as an SSE or HTTP-streamable MCP service and set `advanced_config.timeout` to `300`.
3. Confirm the value took effect (`GET /api/v1/mcp-services/{id}` reports `timeout: 300`), and that connection setup / `tools/list` work.
4. Bind the service to a custom agent and ask a question that calls the slow tool.
5. Inspect `messages.agent_steps` or the backend log for that tool step.

### Expected Behavior

A tool call against an MCP service configured with `timeout: 300` should be allowed to run for the configured duration. `advanced_config.timeout` should mean the same thing at every layer that enforces it.

### Actual Behavior

The tool step fails at ~60000ms with a context deadline, regardless of the service's configured timeout. Ordinary tools and `shell_exec` are unaffected.

### Suggested Fix

Let the per-call budget consult the MCP service behind the tool, falling back to the current name-based default:

- take `max(defaultToolExecTimeout, service.AdvancedConfig.Timeout)` for MCP-backed tools, so a service left at the default `30` keeps today's 60s and nothing regresses;
- apply the same value in the post-approval retry path in `mcp_tool.go`;
- leave built-in tools and the `shell_exec` special case untouched.

This reuses configuration that already exists and is already enforced downstream, rather than adding a new knob. Happy to send a PR with tests if the direction looks right.

### WeKnora Version

Code paths above verified against `main` at `3ab6d394` (2026-09-09). Originally hit on an older self-hosted build while integrating an external long-running analysis MCP service.

### Deployment Method

Docker Compose

### Operating System

Linux self-hosted Docker Compose deployment

### Relevant Logs

Empirically, with the service left at the default `timeout: 30`, a 50s poll call was cut at ~27-30s (transport honoring the config); raising the service to `timeout: 300` fixed that, but individual calls still had to be kept under 60s, which is what forced a `submit` + short-`poll` protocol instead of a single synchronous call.

### Confirmation

- [x] I searched existing issues. #2298 / #2302 cover the same generic 60s cap for the built-in `web_fetch` tool and propose a per-tool default plus an env var; this issue is about MCP tools, where a per-service timeout is already configured by the user and already honored by the transport layer, so no new configuration surface is needed.

LuckVd:
## Description

Fixes #3135

An MCP service's `advanced_config.timeout` (default `30`, `types.GetDefaultAdvancedConfig`) is honored by the transport layers — the MCP HTTP client timeout (`internal/mcp/client.go:165-166`) and the `initialize` handshake (`internal/mcp/manager.go:190-194`) — but never by the agent-side per-call budget. `toolExecutionTimeout` in `internal/agent/const.go:55-60` decides from the tool name alone (`shell_exec` → 10m5s, everything else → 60s), and `internal/agent/act.go:528` applies it to MCP tools too. The post-approval path in `internal/agent/tools/mcp_tool.go` re-derived the same 60s (`meta.ExecTimeout` with a hardcoded 60s fallback). Result: an operator can raise a service to `timeout: 300`, watch the transport honor it, and still have every individual CallTool cancelled at 60s — the configuration is silently partially applied.

### Change

- Add `MCPTool.serviceCallTimeout()` reading `service.AdvancedConfig.Timeout` (0 when unset/non-positive).
- Add `MCPTool.callToolTimeout(engineTimeout)`: the service timeout **extends** the engine-derived window when it is longer; it never shortens it. Widening-only keeps every existing deployment's behavior identical and leaves shorter values enforced where they already apply (the HTTP transport timeout).
- Normal path: after the exec-meta read in `Execute`, re-derive the CallTool window from `meta.ApprovalCtx` (the round-level parent without the per-tool deadline) when the service timeout extends it. The swapped ctx is created after `oauthSess`/`toolCallID` are read, so they keep using the meta-carrying ctx.
- Post-approval path: the existing fresh-window re-derivation now goes through `callToolTimeout`, so approved MCP calls get the configured window too.

### Type of Change

- [x] 🐛 Bug fix

### Related Issue

- Fixes #3135

### Testing

- New tests (`internal/agent/tools/mcp_tool_timeout_test.go`):
  - `TestServiceCallTimeout` — nil service / nil advanced config / zero / negative / positive values.
  - `TestCallToolTimeout` — longer service timeout wins (#3135); equal and shorter never shorten; missing engine timeout falls back to 60s then extends; missing engine timeout with short service keeps 60s.
- `go test ./internal/agent/tools/ -count=1` — pass (full package)
- `go vet ./internal/agent/tools/` — clean
- `gofmt -l` no output; `git diff --check` clean
- Pre-existing issues: none observed in the affected package.

### Checklist

- [x] Self-reviewed the code
- [ ] Breaking changes are clearly called out in the description above → none (widening-only: calls previously succeeding at ≤60s are unaffected)

### DIFF
--- a/internal/agent/tools/mcp_tool.go
+++ b/internal/agent/tools/mcp_tool.go
@@ -109,6 +109,34 @@ func (t *MCPTool) Parameters() json.RawMessage {
 	}`)
 }
 
+// serviceCallTimeout returns the MCP service's configured per-call timeout
+// (advanced_config.timeout, in seconds), or 0 when unset or not positive.
+func (t *MCPTool) serviceCallTimeout() time.Duration {
+	if t.service == nil || t.service.AdvancedConfig == nil || t.service.AdvancedConfig.Timeout <= 0 {
+		return 0
+	}
+	return time.Duration(t.service.AdvancedConfig.Timeout) * time.Second
+}
+
+// callToolTimeout returns the timeout governing the actual MCP CallTool window.
+// The agent engine derives the per-tool budget from a blanket 60s
+// (toolExecutionTimeout in internal/agent), while the service-level
+// advanced_config.timeout was only honored by the transport layers — a service
+// configured with a longer timeout still had every call cancelled at 60s (#3135).
+// The service timeout therefore extends the engine window when it is longer; it
+// never shortens it, so services without an explicit (longer) timeout keep
+// today's behavior and shorter values stay enforced where they already apply
+// (the HTTP transport timeout in internal/mcp/client.go).
+func (t *MCPTool) callToolTimeout(engineTimeout time.Duration) time.Duration {
+	if engineTimeout <= 0 {
+		engineTimeout = 60 * time.Second
+	}
+	if st := t.serviceCallTimeout(); st > engineTimeout {
+		return st
+	}
+	return engineTimeout
+}
+
 // Execute executes the MCP tool
 func (t *MCPTool) Execute(ctx context.Context, args json.RawMessage) (*types.ToolResult, error) {
 	logger.GetLogger(ctx).Infof("Executing MCP tool: %s from service: %s", t.mcpTool.Name, t.service.Name)
@@ -201,12 +229,9 @@ func (t *MCPTool) Execute(ctx context.Context, args json.RawMessage) (*types.Too
 				// Approval may have consumed most/all of the per-tool exec budget set by the
 				// agent engine (act.go). Re-derive a fresh tool-exec ctx from ApprovalCtx so
 				// the actual MCP CallTool gets a full timeout window. (issue #1173 follow-up)
+				// callToolTimeout honors the service's advanced_config.timeout (#3135).
 				if meta.ApprovalCtx != nil {
-					freshTimeout := meta.ExecTimeout
-					if freshTimeout <= 0 {
-						freshTimeout = 60 * time.Second
-					}
-					freshCtx, freshCancel := context.WithTimeout(meta.ApprovalCtx, freshTimeout)
+					freshCtx, freshCancel := context.WithTimeout(meta.ApprovalCtx, t.callToolTimeout(meta.ExecTimeout))
 					defer freshCancel()
 					ctx = freshCtx
 				}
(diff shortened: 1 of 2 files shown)


LuckVd:
Note on the failed `Format, vet, test, and build` check (run 34555136238): the two failures are `TestInstallSkillRetriesAFailedSkillWithTheSameArchive` and `TestInstallSkillReinstallsWhenTheLiveImageNoLongerCarriesTheSkill` in `internal/application/service` (tenant skill-install state machine, asserting status `installing` but got `ready`). This PR only touches `internal/agent/tools` (MCP tool timeout) — a different package with no shared state.

Baseline evidence:
- `internal/agent/tools` passed in the same CI run.
- main@5db13a13 (this PR's base) had a green run of the same workflow.
- Both failing tests pass locally on this exact head sha, 5 consecutive runs.

Looks like a timing-sensitive flake on the CI runner. Re-triggered the workflow with an empty commit to confirm.

LuckVd:
Update: the re-triggered run passed — `Format, vet, test, and build` ✅ 5m51s (run 34557034100), all other checks green. Confirms the earlier failures were a timing-sensitive flake in the two tenant-skill-install tests, unrelated to this change.

Issue #3158, opened 2026-09-10
[Bug]: runtime_context renders the full bound-KB catalog with no size cap — kb=all agents on multi-KB tenants exceed the context window at round 1 and burn tokens in futile rounds


vicfei:
### Affected Component

Backend — agent runtime-context assembly and the context-management loop:

- `internal/agent/prompts.go:130-200` — `formatKnowledgeBaseList`
- `internal/agent/observe.go:510-562` — `buildRuntimeContextBlock`
- `internal/agent/engine.go:642` — `manageContextWindow` orchestration

### Problem

`buildRuntimeContextBlock` renders the bound knowledge bases for the turn into the **current user message**, and `formatKnowledgeBaseList` has **no cap on total size**:

- the `for _, kb := range kbInfos` loop is linear in the number of bound KBs (`prompts.go:137`);
- each KB renders `id/name/type/doc_count/capabilities` + a **full, untruncated `description`** (`prompts.go:148-150`) + recent-document entries — 2 for document KBs (`prompts.go:174-188`) and **up to 10 FAQ entries including full answers** for FAQ KBs (`prompts.go:152-173`).

For an agent whose KB selection is "all knowledge bases" on a tenant with many KBs, the catalog alone exceeds the model's context window **before the conversation has even started**.

### Measured

Synthetic catalog cost on upstream/main (each KB: realistic ~200-char description, 2 recent docs with the capped 120-char summaries the code already applies):

| bound KBs | catalog size |
|---|---|
| 10 | 20.7 KB |
| 50 | 103 KB |
| 100 | 206 KB (~107k chars) |
| 300 | 619 KB |
| 541 | 1.12 MB (~581k chars) |

≈ **2 KB per KB, strictly linear, no ceiling**. 100 KBs is already window-threatening for 128k-context models. Observed on a large multi-KB deployment (541 KBs): the first-turn user message alone estimated at **219,319 tokens against a ~200k window**.

### What happens next (the amplifier)

Every downstream safeguard then behaves *as designed* but none can help, because the oversized input is a birth defect rather than accumulated growth:

1. the token estimator correctly alarms (over the compaction threshold);
2. the compactor correctly declines — there are **0 history messages** and the current turn's user message is exempt by design, so there is nothing it is allowed to compact;
3. **no over-window rejection exists before the model call** — the request goes out anyway;
[...]

PR #3222, opened 2026-09-12
fix(agent): bound the runtime_context KB catalog and fail fast on over-window turns

vicfei:
Implements #3158.

Two independent guards against the unbounded bound-KB catalog described in the issue:

## 1. Catalog budget in `formatKnowledgeBaseList`

Entries render in full detail until `kbCatalogFullDetailChars` (20k) is exhausted, then degrade to
single-line entries that keep every routing id (`knowledge_search` targets bases by these ids), and are
omitted past `kbCatalogHardCeilingChars` (128k) with a note that unscoped searches still cover them.

Measured on this branch with the synthetic corpus from the issue (541 KBs, ~200-char descriptions):

| | before | after |
|---|---|---|
| 541-KB catalog | ~666 KB / 392k chars | ~110 KB, all 541 ids intact |
| 3000-KB catalog | ~3.7 MB | ~135 KB + omission note |

(The per-entry slimming that landed in #3193 already cut entry cost ~40%; this adds the missing total bound
so the catalog can no longer outgrow the window no matter how many bases are bound.)

## 2. Fail-fast guard before the model call

`ensureTurnFitsContextWindow` aborts the turn when the estimated request — messages, plus tool schemas
when no usage baseline exists yet — cannot fit `MaxContextTokens` minus the existing `contextSafetyTokens`
margin. This is the case compaction structurally cannot fix: the oversized payload is the current turn's
own user message, which the compactor correctly refuses to cut. Previously such a turn burned the full
`maxIterations` run resending the same oversized request; now it costs one estimate check and returns an
error naming the two levers an operator actually has: narrow the bound knowledge bases, or configure a
larger-window model.

Unknown/zero `MaxContextTokens` keeps the previous provider-judged behaviour.

## Tests

- `TestFormatKnowledgeBaseList*`: small catalogs unchanged; 541-KB degradation keeps all ids and stays
  bounded; 3000-KB hits the hard ceiling and emits the omission note.
- `TestEnsureTurnFitsContextWindow`: oversized turn rejected with the actionable message; fitting turn
  passes; unknown window no-ops; tool schemas counted when no usage baseline exists.

## Checklist

- [x] `gofmt` clean, `go build ./...` clean, `go vet` clean
- [x] full `internal/agent` package tests green

### DIFF
--- a/internal/agent/context_guard_test.go
+++ b/internal/agent/context_guard_test.go
@@ -1,0 +1,76 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/Tencent/WeKnora/internal/models/chat"
+	"github.com/Tencent/WeKnora/internal/types"
+)
+
+func newGuardTestEngine(t *testing.T, maxContextTokens int) *AgentEngine {
+	t.Helper()
+	return newTestEngine(t, nil, func(cfg *types.AgentConfig) {
+		cfg.MaxContextTokens = maxContextTokens
+	})
+}
+
+// TestEnsureTurnFitsContextWindow pins the fail-fast guard: a current turn
+// that cannot fit the window (no matter what compaction does) aborts with an
+// actionable error instead of resending the same oversized request for the
+// rest of maxIterations.
+func TestEnsureTurnFitsContextWindow(t *testing.T) {
+	var big strings.Builder
+	for big.Len() < 600_000 {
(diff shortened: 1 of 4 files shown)


vicfei:
This was closed without a review or comment — could you share what didn't fit? #3158 is still open, so if the direction is wanted in a different shape I'm happy to rework, e.g.:

- guard placement: fail the turn up front (current) vs. only degrading the catalog and letting the provider judge,
- config-driven limits instead of the fixed budget constants,
- or splitting the two changes (catalog budget / fail-fast guard) into separate PRs.

PR #3249, opened 2026-09-14
fix(session): resolve own-workspace agent when agent_source_tenant_id echoes the caller tenant

### LINKED ISSUE #3161: [Bug]: Own-workspace agent_source_tenant_id breaks chat (404 "Shared agent not found") and KB listing (403)
### Affected Component

Backend Service & API

### Bug Description

A request that carries `agent_source_tenant_id` **equal to the caller's own tenant** is rejected as a missing share, even though the agent belongs to the caller and does not need a share at all.

`agentShareService.GetSharedAgentForTenant` returns `ErrAgentShareNotFound` for that input (`internal/application/service/agent_share.go:465-467`):

```go
if len(sourceTenantID) > 0 && sourceTenantID[0] != 0 {
	if sourceTenantID[0] == tenantID {
		return nil, ErrAgentShareNotFound
	}
```

`session.Handler.resolveAgent` only falls back to the local agent when `sourceTenantID == 0` (`internal/handler/session/qa.go:557-561`), so it returns `nil` and the caller aborts (`internal/handler/session/qa.go:190-193`):

```go
customAgent, effectiveTenantID, sharedAgentReadOnly := h.resolveAgent(ctx, c, request.AgentID, request.AgentSourceTenantID)
if request.AgentSourceTenantID != 0 && customAgent == nil {
	return nil, nil, errors.NewNotFoundError("Shared agent not found")
}
```

This is reachable because an own-workspace agent never has a share row for its own workspace: `GetShareByAgentIDForTenant` deliberately excludes `source_tenant_id == tenantID`. The client keeps that selector together with the conversation (deep link query / restored `last_request_state`) and resends it on every subsequent message, so once the value is set the conversation fails permanently until client-side state is cleared.

### Steps to Reproduce

1. Take an agent owned by tenant `A` — the caller's own agent, no share involved.
2. Send a QA message and include the owning tenant as the source selector:

```bash
curl -X POST "$HOST/api/v1/agent-chat/$SESSION_ID" \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{"query":"hello","agent_id":"<agent-id>","agent_source_tenant_id":<tenant-A-id>}'
```

3. The same selector breaks the @-mention knowledge base list:

```bash
curl "$HOST/api/v1/knowledge-bases?agent_id=<agent-id>&agent_source_tenant_id=<tenant-A-id>" \
  -H "Authorization: Bearer $TOKEN"
```

Note: omitting `agent_source_tenant_id` (or sending `0`) works fine, which is what makes the failure look session-specific.

### Expected Behavior

The caller's own agent runs in the caller's tenant: the chat request succeeds and uses the agent's configured model/KB scope, and the knowledge base list returns that agent's scope.

### Actual Behavior

- `POST /api/v1/agent-chat/:session` → HTTP 404, message `Shared agent not found` (`internal/handler/session/qa.go:191`). The request is rejected before any LLM call, so the conversation cannot proceed at all.
- `GET /api/v1/knowledge-bases?agent_id=…&agent_source_tenant_id=…` → HTTP 403, message `no permission for this shared agent` (`internal/handler/shared_agent_access.go:40`).
- `POST /api/v1/sessions/:id/temporary-documents` shares the same `resolveAgent` guard (`internal/handler/session/temporary_document.go:47-51`), so attachment upload fails the same way.
- Net user-visible effect: sending a message in such a conversation keeps failing (the UI surfaces a generic error, e.g. `code 1002`), the @-mention KB picker is empty, and the agent picker silently falls back to the local builtin.

### WeKnora Version

App Version: v0.8.0 (UI Version: v0.8.0), deployed with Docker Compose.

Also still present on `main` today (`5db13a1`): `internal/application/service/agent_share.go:466` and `internal/handler/session/qa.go:557-561` are unchanged, so the same request fails there too.

### Deployment Method

Docker Compose

### Operating System

Ubuntu 22.04 (Linux, Docker Compose)

### Relevant Logs

```
POST /api/v1/agent-chat/<session_id>
  -> HTTP 404   message: "Shared agent not found"                      (internal/handler/session/qa.go:191)

GET  /api/v1/knowledge-bases?agent_id=<own-agent>&agent_source_tenant_id=<own-tenant>
  -> HTTP 403   message: "no permission for this shared agent"         (internal/handler/shared_agent_access.go:40)

POST /api/v1/sessions/<session_id>/temporary-documents
  -> HTTP 404   message: "Shared agent not found"                      (internal/handler/session/temporary_document.go:49)
```

Removing `agent_source_tenant_id` from the same requests returns 200, which isolates the selector as the trigger.

### Confirmation

vicfei:
Fixes #3161.

## What changes

A request carrying `agent_source_tenant_id` equal to the caller's own tenant was rejected as a missing share even though the agent belongs to the caller and needs no share at all: `GetSharedAgentForTenant` returns `ErrAgentShareNotFound` for that input (`agent_share.go:466-468`), and `resolveAgent` only falls back to the own agent when the selector is absent (`session/qa.go`), so the chat request died with 404.

The fix normalizes the self-referential selector to the absent selector at `resolveAgent` entry via `types.NormalizeAgentSourceTenantID`, so own agents resolve through the existing local path:

- **untouched**: a rejected *foreign* selector still must not silently fall back to a same-ID local agent — that guard is pinned by the existing `TestResolveAgent_RejectsInvalidSharedSelectorWithoutLocalFallback`;
- **new regression test**: `TestResolveAgent_ResolvesOwnAgentWhenSourceSelectorIsSelfTenant` pins the own-agent resolution for the self selector;
- unit tests for the normalization helper cover absent / self / foreign selectors.

## Tests

- [x] `go test ./internal/handler/session/ ./internal/types/` — green
- [x] `gofmt` / `go build ./...` / `go vet` clean

## Note

The same self-selector input reaches two sibling surfaces (`handler/shared_agent_access.go`, `application/access/knowledgebase.go`) where the correct handling is an ownership grant rather than a share lookup; that is a slightly different semantic change, so I kept this PR to the reported chat 404 and can follow up separately if wanted.


lyingbug:
Closing this PR to consolidate the own-workspace agent selector fix into #3162: https://github.com/Tencent/WeKnora/pull/3162. Both address #3161, and #3162 remains open; the bug is not being marked as shipped or resolved.

I compared both patches against current main. This PR normalizes the selector at session.resolveAgent, which fixes the chat entry point. #3162 handles the same selector in GetSharedAgentForTenant and preserves local, writable semantics in resolveAgent. The service is also used by the shared-agent access helper for knowledge-base listing, document search, and batch restoration, so the service-level proposal is the broader integration point for this issue.

Please carry the useful normalization and regression cases from this PR into #3162's review: self selection must resolve locally, absent selection must retain existing behavior, and a rejected foreign selector must never fall back to a same-ID local agent. Handler fallback when the shared-agent service is unavailable also deserves an explicit test; these approaches are not identical in that case.

Thank you for the focused fix and tests. Keeping one open implementation will let the selector behavior be reviewed across all callers.

vicfei:
Understood — consolidating into #3162 makes sense; normalizing at `GetSharedAgentForTenant` covers the non-chat entries that the handler-level fix here did not reach.

One thing from this PR that may be reusable there: the four-state `resolveAgent` tests in `resolve_agent_test.go` (own-selector without local fallback, shared selector match, local fallback, terminal sandbox config) pin exactly the semantics #3162's "preserves local, writable resolution" changes rest on — happy for @lsasw to lift them if useful.

PR #3279, opened 2026-09-15
fix(milvus): keep BM25 scores on keyword hits and rank them before topK

po-et:
## Description

`KeywordsRetrieve` in the Milvus engine threw away the scores Milvus returned (`sets, _, err := convertResultSet(...)`) and stamped every keyword hit with `1.0`. That line dates from the first Milvus integration, before keyword search had a score of its own; with BM25 in place it means every keyword hit ties, and wherever results are ordered by score — the Langfuse `top_hits` preview in #3274, `deduplicateByScore` in keyword-only searches — they sit ahead of every vector hit. The vector path already takes `scores[i]` from the result set; the keyword path now does the same through a small shared helper.

Second thing that fell out of reading it: hits arrive one collection after another and were cut to `topK` in that order, so a better hit in a later collection could be dropped, and RRF fusion — which ranks by position — saw collection order rather than score order. They are now sorted by score (stable, so ties keep Milvus's own order) before truncation.

## Type of Change

- [x] 🐛 Bug fix

## Related Issue

Fixes #3274

## Testing

- `go test ./internal/application/repository/retriever/milvus/` passes; two new unit tests cover the helpers: keyword hits carry the returned scores and match type, and truncation keeps the best hits across collections with ties in arrival order. Reverting the score to `1.0` fails the first, removing the sort fails the second.
- `golangci-lint run --new-from-rev=upstream/main` (v2.12.2, the CI version) reports 0 issues; `go build ./internal/...` and `go vet` on the package pass.
- `go test -count=1 ./...` passes locally, 87 packages, 0 failures.
- No live Milvus was used; the retrieval path itself is exercised by the existing engine setup, and the change is confined to how scores and order are carried out of the result set.

## Checklist

- [x] `git diff --check upstream/main...HEAD` passes
- [x] Changed source files are formatted
- [x] Targeted tests for the changed package pass
- [x] Diff-scoped lint passes
- [x] Full-repository checks were run (see Testing)
- [x] Self-reviewed the code
- [x] Added tests covering the change
- No API, schema, documentation, or UI changes; not a breaking change.

PR #3281, opened 2026-09-15
test(repository): serialise the SQLite pool in the MCP tool approval tests

po-et:
## Description

`TestMCPToolApprovalConcurrentColumnPatches` runs two `UpsertPolicy` calls at once against a shared-cache in-memory SQLite database. When the second call lands on its own pool connection while the first still holds the table, the driver answers `SQLITE_LOCKED` — `upsert mcp tool policy: database table is locked` — which `_busy_timeout` does not cover (that only waits out `SQLITE_BUSY`). Locally on current `main` it fails 4 of 1000 runs, 1 of 100 under `-race`; on CI it failed the "Format, vet, test, and build" lane of #3279, which does not touch this package.

The package already has the answer in two sibling helpers (`knowledge_create_test.go`, `knowledge_finalize_test.go`): cap the pool at one connection. The two upserts then serialise on the pool and the test still exercises what it was written for — the two column patches must not clobber each other. 0 of 1000 and 0 of 100 under `-race` afterwards.

## Type of Change

- [x] 🧪 Test

## Related Issue

None; seen on the CI run for #3279.

## Testing

- `go test -count=1000 -run 'TestMCPToolApprovalConcurrentColumnPatches$' ./internal/application/repository/`: 4 failures before, 0 after; `-race -count=100`: 1 before, 0 after.
- `go test ./internal/application/repository/` passes; `golangci-lint run --new-from-rev=upstream/main` (v2.12.2, the CI version) reports 0 issues.
- Test-only change; production code untouched.

## Checklist

- [x] `git diff --check upstream/main...HEAD` passes
- [x] Targeted tests for the changed package pass
- [x] Diff-scoped lint passes
- [x] Self-reviewed the code
- [x] Test-only change
- No API, schema, documentation, or UI changes.


po-et:
ac4dd597 on main already applies the same SetMaxOpenConns(1) to this test, with a per-test UUID DSN and a start barrier on top, so this one is redundant — closing.

Issue #2873, opened 2026-08-30
[Feature]: Add local filesystem/folder datasource connector for continuous local directory sync


Saolence:
## Affected Component

Backend Service & API / Document Reader Service / Data Source Framework

## Problem Description

WeKnora currently has **no way to keep a knowledge base continuously in sync with a local filesystem folder**. The data source framework only ships connectors for remote services. From `internal/container/container.go` (`initConnectorRegistry`), the registered connectors are: `feishu`, `lark`, `feishu_drive`, `lark_drive`, `notion`, `yuque`, `ima`, `rss`, `gitlab` — all remote. There is no `local_folder` / filesystem connector in `internal/datasource/connector/`, and there is no file-watch (fsnotify/inotify) mechanism anywhere in the codebase.

For self-hosted deployments (the "one laptop to Kubernetes" spectrum described in the docs), the local disk is often the primary place documents live. Right now a user must either re-upload files manually or build an external sync script. The missing primitive is "bind a local directory to a knowledge base, keep it current automatically".

There is also a structural blocker for doing this with the existing upload API: `internal/application/service/knowledge_create.go` deduplicates uploads by **fileName + fileSize + fileHash**. If a local file is *modified* (new hash) and re-uploaded, WeKnora creates a **second** document instead of updating the existing one — duplicates accumulate on every change. `PUT /knowledge/:id` (`cli/cmd/doc/update.go`) only mutates title/description, not file content, so there is currently no way to "update the content of an existing knowledge entry".

## Proposed Solution

Add a **Local Filesystem connector** to the data source framework (type `local_folder`), reusing the existing Connector/Scheduler/SyncLog infrastructure:

1. **`LocalFolderConnector`** implementing `internal/datasource/connector.go`'s `Connector` interface:
   - `Type()` → `local_folder`
   - `ListResources` → subdirectories of the configured root
   - `FetchAll` → walk the selected directory, emit every file as a `FetchedItem`
   - `FetchIncremental` → walk and compare `mtime`/size against the stored `SyncCursor` (last sync time), emitting only changed/new files; deleted files are reported so `SyncDeletions` can remove the corresponding knowledge entries
[...]

PR #3293, opened 2026-09-15
feat(knowledge): add in-place file replacement primitive

mazjq:
## Summary

Data sources that re-sync an existing file knowledge currently have to delete it and
create a new one, so the knowledge ID changes on every content update. This PR adds a
general `KnowledgeService.ReplaceKnowledgeFile` primitive that replaces the stored source
file of an existing file knowledge in place, preserving its knowledge ID, and re-parses it
through the existing `ReparseKnowledge` pipeline.

## Why

Connectors that sync continuously need a stable knowledge identity, so that citations,
retrieval results, tags and anything else referencing a knowledge ID survive a content
update instead of pointing at a deleted row. Delete + recreate also discards the history of
what is, from the user's point of view, still the same document.

This PR does not change the behavior of existing connectors. It only adds a new method; no
existing API, handler or connector calls it yet.

It is groundwork for #2873 (local folder data source), where a directory is re-scanned on a
schedule and an edited file must keep its knowledge ID.

## Implementation

Object storage is not transactional, so the steps are ordered to keep the knowledge row
pointing at a file that exists:

- save the new source file
- update the source metadata in one UPDATE, preserving the knowledge ID (stored metadata is
  merged, not overwritten)
- re-parse through the existing `ReparseKnowledge` pipeline
- delete the old source file only after the update and re-parse succeed
- compensate on failure: a failed column update discards the new file; a failed re-parse
  restores the previous source columns and discards the new file. The row is re-read before
  a replacement file is deleted, so a write that committed despite reporting an error never
  leaves a knowledge referencing a deleted file.

Identical content under the same name and folder is not re-parsed: changed metadata is
persisted and a `DuplicateKnowledgeError` is returned, matching `CreateKnowledgeFromFile`'s
duplicate contract.

## Testing

Run on this branch (Go 1.26, Linux):

- `go test -run ReplaceKnowledgeFile ./internal/application/service/` — 9 tests covering ID
  preservation, the unchanged-content skip, each failure and compensation path, and input
  validation
- `go test ./internal/application/service/ ./internal/types/...`
- `go build ./...` and `go build ./cmd/server`
- `go vet` on all non-docreader packages
- `golangci-lint` v2.12.2: no new issues introduced by this commit relative to its base;
  `gofmt` clean on the changed files

## Follow-up

A follow-up PR will add the Local Folder data source for #2873, the first consumer of this
primitive.


lyingbug:
Superseded by https://github.com/Tencent/WeKnora/pull/3303.

That PR keeps this primitive (thanks @mazjq) and adds the review follow-ups: dequeue in-flight parse tasks before switching `file_path`, mark pending in the same UPDATE, skip stale `ProcessDocument` Saves when the stored path has moved, and keep the existing folder unless `customFileName` is path-qualified.

PR #3294, opened 2026-09-16
feat(chat): show agent context usage beside the model selector

nullptr-error:
## Summary
- Attribute the last agent LLM prompt into system / tools / conversation / MCP / skills and show it as a ring next to the model selector, so you can see how the context window is spent.
- Refresh the ring after each calibrated LLM round (not only on complete), without writing mid-turn usage onto the in-memory message (that raced the stop path) or flashing Tools/MCP to zero during synthesis.
- Overlay the composer on a full-height thread with a darker scrollbar thumb that still reaches the bottom and tracks the right panel.

## Test plan
- [ ] Start an agent turn and confirm the ring appears after the first LLM round, then updates each later round (one jump per round, not estimate then measured).
- [ ] New-session / empty composer: no 0% green ring until there is a snapshot. Embedded / quick Q&A: ring stays hidden.
- [ ] Stop mid-turn: persisted assistant `usage` stays NULL; completed turns still store the last snapshot.
- [ ] A turn that synthesizes a final answer: Tools/MCP do not flash to 0 while the answer streams; the complete snapshot may still show 0 for those buckets (last request has `ToolChoice=none`).
- [ ] Widen the right panel: composer stays aligned; scrollbar thumb can be dragged to the bottom; messages clip at the composer top.
- [ ] Reload history / GET messages: usage JSON has no empty `"context":{}`.


lyingbug (commented):
Thanks — a context-usage ring next to the model selector is something we'd like to have, and the care around not racing the stop path shows. Main has moved a lot in this area since you opened it, though, so it needs a rebase plus a couple of alignment changes:

1. **Rebase onto main.** It currently conflicts in `internal/agent/engine.go`, `internal/agent/finalize.go`, `internal/types/chat.go`, `frontend/src/views/chat/index.vue` and the `docs/` swagger files, mainly because of #3413 (window-sized history + token calibration), #3449 and #3549. In particular `turnUsage()` in `finalize.go` now also returns usage when only `ContextTokenScale` is set, so please merge your `Context` condition into that rather than replacing it.
2. **Reuse the existing window helper.** `contextWindowTokens()` in `engine.go` is the same logic as `HistoryTokenBudget(cfg)` in `internal/agent/const.go` (~178). Please call that (or pull one shared helper out of it) so the ring and the history loader can't end up with different windows.
3. **Keep one calibration.** After #3413, `e.tokenEstimator` is already scaled by the learned `ContextTokenScale` (`internal/agent/calibration.go`), and that code deliberately calibrates *conversation only*, because providers render tool schemas differently. `ContextUsage.Calibrate` then scales every bucket, tools and MCP included, by the same ratio on top of that. Please build the attribution on main's calibration instead, e.g. estimate tools/system with `rawEstimator`, price conversation with the scaled estimator, and if you still want the total to equal `prompt_tokens`, put the leftover into one clearly named bucket instead of scaling all of them.
4. **Drop the regenerated `docs/docs.go`, `docs/swagger.json` and `docs/swagger.yaml`** from this PR. Three of the conflicts are in these files, and the generated swagger under `docs/` is currently out of date anyway. If the new `usage.context` field needs documenting, `website-docs/` is the place for it.

PR #3310, opened 2026-09-16
fix(agent): honor JSON Schema type unions in tool parameter handling

lvchenyang:
## Description

Tool parameters declared as a Go slice, map or pointer lose both type coercion and schema validation, because their generated schema declares a **type union** that `CastParams` and `validateProperty` do not recognise.

`jsonschema.For` (google/jsonschema-go v0.4, `go.mod:29`) emits nilable Go types as `{"type":["null","array"]}` rather than `{"type":"array"}`. Both helpers read the declared type with `prop["type"].(string)`, which yields `""` for a union and takes the `continue` / no-op branch — the parameter is passed through untouched and unchecked.

The user-visible failure is an agent loop that cannot make progress. A model that stringifies an array argument (routinely seen with Qwen, and the exact reason `castValue`'s `"array"` branch exists) sends:

```json
{"task":"...","steps":"[{\"id\": \"s1\", \"description\": \"...\", \"status\": \"in_progress\"}]"}
```

`CastParams` leaves `steps` as a string, `ValidateParams` raises nothing, and `TodoWriteTool.Execute` fails to decode:

```
[Agent][Round-6][Tool todo_write (1/?)] Failed in 3ms: json: cannot unmarshal string
into Go struct field TodoWriteInput.steps of type []tools.PlanStep
```

The model retries the identical call each round, so the session burns rounds until it hits the iteration cap. `todo_write` is only the most visible victim — `knowledge_search`, `get_document_info`, `query_knowledge_graph`, `web_fetch`, `data_analysis` and the `wiki_*` tools all declare slice parameters and are equally unprotected.

Fix: resolve the declared type through a shared `schemaType` helper that accepts either a string or a union, picking the first non-`null` member. Both call sites use it, so casting and validation come back for every nilable parameter.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue

N/A — found while running an agent session against Qwen; no existing issue tracks it.

## Testing

Reproduction, against the schema `jsonschema.For[TodoWriteInput]` actually produces on v0.4.3:

```
schema  {"type":"object","properties":{"steps":{"type":["null","array"],...}},...}
args    {"steps":"[{\"id\":\"s1\",\"description\":\"a\",\"status\":\"in_progress\"}]","task":"t"}

before  CastParams -> {"steps":"[{\"id\":\"s1\",...}]","task":"t"}   (unchanged)
        json.Unmarshal -> cannot unmarshal string into Go struct field
                          TodoWriteInput.steps of type []PlanStep
after   CastParams -> {"steps":[{"description":"a","id":"s1","status":"in_progress"}],"task":"t"}
        json.Unmarshal -> ok
```

`TestCastParams_NullableArrayTypeUnion` covers that shape. Verified it fails on `main` and passes on this branch:

```
$ go test ./internal/agent/tools/ -run TestCastParams_NullableArrayTypeUnion   # with main's param_cast.go
--- FAIL: TestCastParams_NullableArrayTypeUnion (0.00s)
    param_cast_test.go:119: expected steps to be array, got string
```

Commands run:

- `go test ./internal/agent/tools/ -run 'TestCastParams|TestValidateParams' -v` — ok, 22 tests pass
- `go test ./internal/agent/tools/ -skip 'TestLoadFromExcel_MultiSheet|TestLoadFromExcel_SingleSheet|TestLoadFromExcel_QuotedSheetName'` — `ok ... 4.122s`
- `gofmt -l` on the three changed files — clean
- `git diff --check origin/main...HEAD` — clean

The unskipped package run times out after 600s on this machine, hanging in `newTestDuckDB` (`data_analysis_duckdb_test.go:28`) before any assertion — a DuckDB setup problem in this environment, not reachable from the changed code. `golangci-lint` is not installed here, so the diff-scoped lint was not run.

## Checklist

- [x] Targeted tests for the changed packages/components pass (DuckDB-dependent cases excluded, see above)
- [x] Self-reviewed the code
- [ ] Updated related documentation (README, `docs/`, Swagger annotations, etc.) — not applicable, internal helper
- [x] Breaking changes are clearly called out in the description above (none)

## Screenshots / Recordings

Not applicable — no user-visible UI change.


lyingbug:
`schemaType` currently returns only the first non-null member of a type union, and `ValidateParams` now treats that member as the only allowed type. This introduces a regression for valid unions with more than one concrete type.

Minimal reproduction against head `9a93f23f5eb5d50d4bdfd84830aa09a56f1f57ab`:

```go
schema := json.RawMessage(`{"type":"object","properties":{"value":{"type":["string","object"]}}}`)
args := json.RawMessage(`{"value":{"nested":true}}`)
errs := ValidateParams(CastParams(args, schema), schema)
// [{value parameter 'value' should be type 'string'}]
```

I ran this against the PR's parameter casting/validation files and current main: the valid object is rejected on this PR and accepted on main. Reordering union members should not change which JSON values are accepted.

Please preserve all allowed types during validation and avoid coercing a value that already matches one of them. If this change is intentionally limited to nullable single-type schemas, only use the scalar-type shortcut when exactly one non-null type remains. Add regression coverage for both nullable arrays and multiple concrete members, including reversed member order.

PR #3326, opened 2026-09-16
fix(model): preserve extra config in connection tests and updates

xyzizz:
## Description

Preserve model `extra_config` across connection tests and updates.

- Send draft `thinking_control` for remote chat connection tests.
- Merge stored provider config per key without overwriting draft values.
- Do not inherit old config after a provider change.

## Type of Change

- [x] 🐛 Bug fix
- [x] 🧪 Test

## Testing

- `npm test -- src/components/modelEditorFlow.test.ts` — 21 tests passed
- `npm run type-check` — passed


lyingbug:
The current frontend CI failure comes from the test added in this PR, so it needs to be fixed before merging.

[Failing job](https://github.com/Tencent/WeKnora/actions/runs/35122030598/job/105085033890): `remote chat connection tests send the draft thinking config`, at `frontend/src/components/modelEditorFlow.test.ts:119`. The reported actual and expected values both contain `{ thinking_control: 'enable_thinking' }`, but `deepStrictEqual` rejects them.

The fixture evaluates the component with `runInNewContext`, so the actual `extraConfig` object comes from a different VM realm. With `node:assert/strict`, `assert.deepEqual` also compares prototypes, causing this assertion to fail despite matching payload fields.

Please compare the relevant scalar property (and keys if exact shape matters), or normalize the payload into the test realm before the deep comparison, then rerun the frontend checks. This is a test-harness issue; the CI output alone does not establish a defect in the model configuration behavior.

PR #3331, opened 2026-09-16
feat(datasource): add local folder connector with in-place incremental sync

### LINKED ISSUE #2873: [Feature]: Add local filesystem/folder datasource connector for continuous local directory sync

mazjq:
## Summary

Adds a `local_folder` data source that keeps a knowledge base in sync with a
directory the WeKnora server can read — typically a bind-mounted Obsidian vault
(#2873). It builds on `KnowledgeService.ReplaceKnowledgeFile`, merged in #3303,
so an edited file keeps its knowledge ID instead of being deleted and recreated.

## How it works

- Scheduler-driven full and incremental scans. mtime and size decide whether a
  file is read at all; SHA-256 confirms a real change; per-file state lives in
  the data source cursor, so incremental sync survives restarts.
- Existing files are updated in place through `ReplaceKnowledgeFile` via an
  opt-in `FetchedItem.UpdateInPlace` flag. Other connectors keep delete+recreate.
- A file that becomes empty keeps its knowledge ID while its old chunks and
  index are cleared; new empty files are not ingested.
- Identity is `local_folder:<relative path>`, scoped per data source, so
  same-named files in different folders stay distinct.
- Include/exclude doublestar patterns, extensions matched case-insensitively.
  The default include covers common document formats; images and audio need
  VLM/ASR and must be added explicitly. Extensions the import pipeline does not
  support are skipped silently rather than failing every sync.
- Scheduled syncs defer files modified in the last 10s (still being written);
  a manual "sync now" does not. `FetchAllFromCursor` lets full syncs reconcile
  deletions too.

## Deployment and safety

- Disabled unless `WEKNORA_LOCAL_FOLDER_ROOTS` is set; that variable is the
  security boundary for what a data source may read.
- `root_path` must be absolute and, after symlink resolution, stay inside an
  allowed root. Walks and reads go through `os.Root`; symlinks and non-regular
  files are skipped.
- An inaccessible root or a failed scan always fails the sync instead of
  deleting knowledge. When a previously synced root suddenly scans empty, a
  scheduled sync fails rather than deleting everything — that is
  indistinguishable from a broken mount — while a manual "sync now" treats it as
  the user confirming the folder's current state and reconciles it (deletions
  still honour `sync_deletions`).
- `root_path` is a filesystem path visible to the WeKnora backend process.
  Docker/Kubernetes deployments require mounting that path into the
  container/pod; Lite, Homebrew, Desktop, and source builds can use paths
  directly visible on the host. Mounting the source read-only (`:ro`) is
  recommended since the connector never writes to it.
- For Kubernetes, the source directory must be mounted into the app pod (for
  example via PVC/NFS/CSI). The current Helm chart does not expose generic extra
  volume mounts, so this requires a post-render/custom chart today.

## Storage note

This connector reuses the existing ingestion path and `FileService`, so files
are **copied into WeKnora-managed storage** rather than referenced in place. The
source directory stays the source of truth, but the content exists twice, plus
chunk/vector/index overhead. Documented in the data source docs; a zero-copy or
reference mode would be a separate change.

## Testing

- `go test ./internal/datasource/... ./internal/application/service/ ./internal/types/... ./internal/container/ ./internal/application/repository/`
- 22 connector tests (scan/exclude/quiet period/case-insensitive extensions,
  cursor restart, symlink and allowlist escapes, mass-deletion guards, manual
  reconciliation of an emptied folder, empty files, oversized files) and service
  tests for the in-place path
- `go build ./...`, `go vet`, `gofmt`; `golangci-lint` v2.12.2 reports no new
  issues relative to the base commit
- Frontend `type-check` and `build`; `npm test` matches the baseline on main
- End-to-end against a running instance: first scan, exclusions, edit keeps the
  knowledge ID with chunks and hybrid search updated, touch is a no-op, create,
  delete with `sync_deletions` off then on, emptied file, and cursor validity
  across a restart
- Deployment check with an ordinary host folder bind-mounted read-only, driving
  create/modify/delete from the host side, plus a UI pass through the dialog
  (test connection, save, sync now, edit hydration)

## Known limitations

- The cursor advances even when an ingest fails (shared semantics of the batch
  sync path); a full sync re-submits those files.
- A rename is a delete plus a create, so the knowledge ID changes.
- Deletions that happen while `sync_deletions` is off are not replayed later.
- No streaming: a file's content is read into memory during sync.
- In multi-tenant deployments, the allowed roots are an operator decision.

Closes #2873


lyingbug (commented):
Thanks, this is a feature we want (#2873), and the safety design is solid: the allowed-roots boundary, `os.Root`, the mass-deletion guard, and in-place updates through `ReplaceKnowledgeFile`. There is one blocker before we can merge:

1. **Unbounded memory per sync.** The scan loop in `internal/datasource/connector/localfolder/connector.go` calls `root.ReadFile` for every pending file and appends the bytes to one `[]types.FetchedItem`. `ProcessSync` only starts ingesting after that slice has been returned. So peak memory is the total size of all new or changed files. `MAX_FILE_SIZE_MB` limits each file, not the total. The first import of a large vault (or a full sync, since `FetchAllFromCursor` re-emits every file) can OOM the app container. Your doc note in `website-docs/03-features/10-datasource.md` describes this, but documenting it doesn't make it safe as a default. Please do one of the following:
   - **Preferred:** implement `datasource.StreamingConnector` / `FullStreamingConnector` (`internal/datasource/connector.go`). The service already has `processSyncStreaming` (fetch → ingest → checkpoint), which keeps memory bounded and resumes after a timeout (#2136). Read each file inside the handler callback, and advance the cursor per file.
   - **Minimum:** cap each sync by total bytes and/or file count. Stop emitting once the cap is hit, and don't advance the cursor for the files you skipped, so the next run picks them up. Make the cap configurable, and document it next to `WEKNORA_LOCAL_FOLDER_ROOTS`.

   Either way, please add a test that a sync over the cap (or an interrupted streamed sync) picks up the remaining files on the next run.

Everything else looks good to me.

mazjq:
Thanks for the review. I took the preferred approach and updated the connector to implement StreamingConnector and FullStreamingConnector.

Sync now processes one file at a time — read → Emit → checkpoint — instead of accumulating all changed files in a single []FetchedItem. File-content memory is therefore bounded to one file at a time, independent of the total folder size. Interrupted or timed-out runs resume from the last persisted checkpoint instead of starting over.

For full syncs, the previous file map is kept in the cursor as a deletion baseline while the new file map is rebuilt. This lets an interrupted full sync retain both its progress and a safe basis for deletion reconciliation. Files deferred by the quiet period or skipped because they exceed the size limit retain their prior cursor state and are not mistaken for deletions.

I also added the requested interruption/resume coverage: an interrupted stream resumes from the last checkpoint, skipping the files already completed and picking up the remaining ones on the next run; an interrupted full stream retains its deletion baseline; and cancellation stops at the last safe checkpoint. There are now 28 local-folder connector tests in total.

CI is green on the rebased branch.

Issue #3338, opened 2026-09-17
[Bug]: async document delete can still leave a stuck row that blocks re-upload ("document already exists") and inflates the KB count


vicfei:
### Affected Component

Backend Service & API

### Bug Description

A user on **v0.8.0** reported this sequence in the community group (translated):

1. Upload a document → UI shows *parse failed*.
2. Delete the document — appears to succeed.
3. The KB then says it holds **4 documents while only 3 are listed**.
4. Re-uploading the same file is rejected with *"document already exists"*.
5. Only manually deleting the row in the database recovered.

This is **after** #1699 (dead-letter marking on retry exhaustion) and #2195 (stop async delete from leaving visible/zombie rows) — both are in v0.8.0 — so it looks like a residual path those fixes don't cover. I verified the following on current `main`:

**The duplicate check only excludes `failed`.** `CheckKnowledgeExists` (`repository/knowledge.go`) filters `parse_status <> 'failed'`, so a row stuck at `processing` **or `deleting`** blocks re-upload with "already exists". A parse pipeline that stalls before writing `failed` (e.g. the unreadable-model stall fixed in #3333, or a worker death mid-parse) produces exactly that.

**The KB count includes the ghost.** `CountKnowledgeByKnowledgeBaseID` has no `parse_status` filter, so the stuck row keeps counting while views that filter by status/folder show fewer — the "4 vs 3".

**Delete is async and its failure modes can strand a row.** The UI submits the delete task and toasts "submitted" (`KnowledgeBase.vue` `deleteSubmitted` + polling); the worker marks rows `deleting` only when the task actually runs (`knowledge_delete.go` "Mark all as deleting first"). If the worker dies mid-execute the row sits at `deleting` until retries exhaust, and the dead-letter callback only re-marks rows that are **still actively deleting** — `UpdateActiveDeletingKnowledgeColumns` skips rows whose status moved on in the meantime (e.g. a *re-parse* click during the window), leaving them `pending`/`processing` forever. Both states block re-upload (see above) and both keep counting.

### Steps to Reproduce (mechanism-level)

- Produce a knowledge row that never reaches `parse_status = 'failed'` (stalled pipeline, or worker killed mid-parse), then delete + re-upload the same file → duplicate error.
- Or: let a delete task's retries exhaust while the row's status was moved out of `deleting` in between (re-parse race) → row survives non-failed, invisible to the dead-letter fix.

### Expected Behavior

[...]

PR #3339, opened 2026-09-17
fix(knowledge): stop deleting rows from blocking duplicate uploads

vicfei:
Refs #3338 — first slice: unblock the user-facing dead end (re-upload rejected by a stranded deleting row). The orphan sweep proposed in the issue is the deeper follow-up.

## Problem

`CheckKnowledgeExists` excluded only `parse_status = 'failed'`. A row whose async deletion was in flight — or stranded (worker death mid-delete; or the dead-letter re-mark skipped because the status had moved out of `deleting` in the meantime) — kept the duplicate check matching forever. Re-uploading the same file then returned **"document already exists"** with no path back except manual SQL, exactly as reported in #3338 on v0.8.0.

## Fix

Exclude `deleting` alongside `failed` in the duplicate check's base filter.

No staleness threshold is needed: a deleting row is on its way out, so an upload landing while the delete task is still queued or running ends with **exactly one live row whichever way the task concludes** —

- task succeeds → the old row is soft-deleted, the new upload remains;
- task exhausts retries → the old row is marked `failed` (which never blocked a retry anyway);
- task is stranded → the old row stays `deleting` but no longer blocks re-upload.

The race that motivated keeping `deleting` blocking does not exist: the delete task plans by row ID, so it can never swallow the newly uploaded row.

## Tests

New `TestCheckKnowledgeExists_ParseStatusMatrix` pins the policy — `failed`/`deleting` allow the upload, `pending`/`processing`/`completed` remain duplicates — on both the hash path and the filename/size fallback. Package green.

## Out of scope (follow-ups from the issue)

- Orphan sweep for rows stuck in `deleting` with no live task (count inflation, auto-recovery).
- The duplicate-hit response already returns the existing knowledge object (`handleDuplicateKnowledgeError` `data`), so a client-side "remove the stuck entry" affordance needs no backend change.


lyingbug:
Landed as the first slice of #3338 (duplicate check no longer matches `deleting`).

Follow-up for the leftovers called out as out of scope here — count inflation, orphan sweep for stranded `deleting`, client duplicate-hit affordance — is #3345.

PR #3360, opened 2026-09-17
fix(knowledge): reject invalid JSON uploads before async parse

xiaojunxiang2023:
## Description

`.json` uploads previously only checked the file extension. Malformed payloads (JSONC comments, leading junk, trailing commas) were accepted, then failed later in async document processing with `DOCREADER_PARSE_FAILED` after retries.

This change validates JSON content at upload and knowledge-replace time using the same gate as the simple JSON reader (`docparser.ValidateJSONContent`), returning HTTP 400 immediately.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test

## Related Issue
N/A — discovered while importing a JSONC-style config file into a knowledge base.

## Testing
- [x] `go test ./internal/infrastructure/docparser/ -count=1 -run 'ValidateJSONContent|SimpleFormatReader_JSON'` — passed
- [x] `go test ./internal/application/service/ -count=1 -run 'ValidateJSONUploadContent'` — passed
- [x] `git diff --check upstream/main...HEAD` — passed

Reproduction / verification:
1. Upload a `.json` file that contains `//` comments or a non-JSON leading line.
2. Before: upload succeeds, parse eventually fails with `invalid JSON content`.
3. After: upload is rejected with HTTP 400 `invalid JSON content` (or `empty JSON content`).

## Checklist
- [x] Self-reviewed the code

## Screenshots / Recordings
N/A (API behavior only)


lyingbug (commented):
Thanks, failing a malformed `.json` upload with a 400 up front beats making users wait for `DOCREADER_PARSE_FAILED` after the retries, and reusing one gate for the reader and the upload is the right shape. A few things before merging:

1. **Datasource sync regression.** `CreateKnowledgeFromFile` isn't only called by the upload handler. `datasource_service.go` (~L1342) calls it on sync, and on an update it has already deleted the existing knowledge for that `external_id` before calling. Before this PR, a bad JSON item still produced a visible document in the failed state. With the check in the service, the old doc is deleted, nothing replaces it, and the sync just logs an error. The IM path (`internal/im/service.go` ~L3442) also swallows the error silently. Please either run the check only in the HTTP upload/replace handlers, or add an explicit option so the datasource path keeps its current behavior, and add a test for whichever you choose.
2. **`readMultipartFileContent` in `knowledge_util.go`: drop the `f.Seek(0, io.SeekStart)`.** Every `file.Open()` returns a fresh reader (`calculateFileHash` and `SaveFile` open their own), and this one is closed right after, so the rewind does nothing. The comment describing it is misleading too. That leaves the helper as `Open` + `ReadAll`.

Optional / non-blocking:
- Messages in this area of `knowledge_create.go` are mostly user-facing Chinese (e.g. "暂不支持上传视频文件"). Since "invalid JSON content" / "empty JSON content" now go straight to the UI as a 400, consider wrapping them in a friendlier message such as "JSON 文件格式不合法：…", and leave the parser's own error unchanged.

xiaojunxiang2023:
@lyingbug, Thanks for the review!

1. Agreed on the datasource sync regression. The JSON gate now runs only in the HTTP upload handler and in `ReplaceKnowledgeFile`. `CreateKnowledgeFromFile` itself no longer validates content, so delete-then-create sync and IM still get a knowledge row that fails in async parse. Covered by `TestCreateKnowledgeFromFileAcceptsInvalidJSONForAsyncParse`.

2. Dropped the `Seek` in `readMultipartFileContent` (and the misleading comment) — each `Open()` already returns a fresh reader.

On the optional friendlier Chinese wrap: left the parser messages as-is for now so the upload 400 stays aligned with the reader error text. Happy to follow up if you'd prefer the UI-facing prefix.

Force-pushed in 1021fef.

PR #3375, opened 2026-09-17
fix(repository): exclude deleting rows from knowledge base count

### LINKED ISSUE #3345: [Bug]: stranded deleting rows still inflate KB count and never auto-recover (#3339 follow-up)
### Affected Component

Backend Service & API (count + housekeeping); Frontend UI (duplicate-hit affordance)

### Bug Description

Follow-up to #3338 after #3339.

#3339 stopped `parse_status = 'deleting'` from matching `CheckKnowledgeExists`, so a stranded async delete can no longer trap the user in a permanent **"document already exists"** with no path back except SQL. That was the first slice; it does **not** close #3338.

The original v0.8.0 report still has two remaining symptoms, and #3339 makes the count one slightly worse if the user actually re-uploads:

1. **KB card still counts the ghost.** `CountKnowledgeByKnowledgeBaseID` has no `parse_status` filter, while the default document list and `ListKnowledgeFolderCounts` already hide `deleting`. Same row → **"4 documents, 3 listed"**. After a successful re-upload on top of a stranded delete: **5 vs 4**.
2. **A `deleting` row with no live task is never recovered.** Housekeeping only sweeps `pending` / `processing` / `finalizing`. The delete dead-letter callback (`markKnowledgeListDeleteFailed`) only flips rows that are **still** `deleting` (`UpdateActiveDeletingKnowledgeColumns`). Worker death mid-delete, or a status that moved off `deleting` before retries exhausted (the re-parse race in #3338), leaves a hidden row forever: storage not released, count inflated, no failed/retry card.

#3339's "exactly one live row whichever way the delete task concludes" also does not hold:

- task succeeds → old row soft-deleted, new upload remains (one row);
- retries exhaust → old row becomes visible `failed` **plus** the new upload (two rows);
- task is stranded → hidden `deleting` **plus** the new upload (two live rows, list still lies about the count).

The ID-scoped delete plan is still correct: it cannot swallow the new upload. The leftover is cleanup, not the duplicate check.

### Steps to Reproduce (mechanism-level)

Confirm on current `main` (includes #3339):

```sql
SELECT id, file_name, parse_status, deleted_at, updated_at
FROM knowledges
WHERE knowledge_base_id = '<kb-id>' AND deleted_at IS NULL;
```

- A ghost with `parse_status = 'deleting'` is hidden from the list, still included in `knowledge_count`, and **no longer** blocks re-upload.
- Housekeeping (`internal/application/service/knowledge_housekeeping.go` Sweep A) will not promote it to `failed`.
- Re-uploading the same file now succeeds and inserts a second row; the ghost remains until manual SQL or a future sweep.

The `processing` stall called out in #3338 is a weaker leftover: housekeeping already recovers `pending`/`processing`/`finalizing` with no live task. Duplicate check still treats a not-yet-swept `processing` row as a hit — see slice 3.

### Expected Behavior

- `knowledge_count` on the KB card matches what the document list shows (in-flight / stranded `deleting` rows must not inflate it).
- A `deleting` row whose delete task is gone (no queued/active `knowledge:list_delete` for that ID) is finished idempotently, or marked `failed` with a clear error so it becomes visible and actionable — same contract housekeeping already gives stalled parses.
- Re-upload hitting a still-blocking row (`processing`, `cancelled`, …) should surface `data.id` / `data.parse_status` so the client can offer "remove the stuck entry", not only a generic "file already exists".

### Suggested slices

1. **Count alignment (small).** Exclude `deleting` from `CountKnowledgeByKnowledgeBaseID` the same way `applyKnowledgeListFilter` / `ListKnowledgeFolderCounts` already do. Pins the original "4 vs 3" report without waiting on a sweep. Add a repository test; there is currently no coverage of this helper.
2. **Orphan sweep for `deleting`.** Extend housekeeping (or a sibling pass) to find `deleting` rows with no live asynq delete task / no durable pending op, then either re-run `DeleteKnowledgeList` idempotently for that ID or mark `failed` via the existing dead-letter column update. Do **not** treat a still-queued delete as orphaned (`filterOutQueued` is the template). This is also the auto-recovery for the dead-letter skip path (status moved off `deleting` before exhaustion — those land in Sweep A if they became `pending`/`processing`; true leftovers are the ones that stayed `deleting`).
3. **Duplicate-hit UX (frontend).** `handleDuplicateKnowledgeError` already returns the existing knowledge in `data`. `KnowledgeBase.vue` / `useKnowledgeBase.ts` currently ignore it and toast `knowledgeBase.fileExists`. Offer delete-then-retry when `parse_status` is a stuck non-terminal state. No backend change required for this slice.

Slice 1 is independently shippable and is the remaining user-visible half of the community report. Slice 2 is the actual leftover-row fix. Slice 3 only matters for statuses #3339 did not exclude.

### WeKnora Version

`main` after #3339 (merged 2026-09-17). Original report was 0.8.0; mechanism of the leftovers verified against current `CountKnowledgeByKnowledgeBaseID` / housekeeping Sweep A / `markKnowledgeListDeleteFailed`.

### Confirmation

- [x] I have searched existing issues (#3338, #2192 / #2195, #1699) and confirmed this is the leftover after #3339, not a duplicate of the duplicate-check fix.

alter:
Closes #3345

### Root cause

A knowledge row that's mid-deletion is marked `parse_status = 'deleting'` until the delete
finishes. `ListKnowledgeFolderCounts` (`internal/application/repository/knowledge.go`)
already excludes those rows from its per-folder counts, but
`CountKnowledgeByKnowledgeBaseID` — the query that feeds `KnowledgeBase.KnowledgeCount`, shown
to users as "how many documents are in this knowledge base" — ran the same shape of query
without that filter. A row stranded in `deleting` state (for example if the delete job never
completes) keeps inflating the displayed count indefinitely.

### Fix

Added the same `parse_status <> 'deleting'` condition to `CountKnowledgeByKnowledgeBaseID`'s
WHERE clause, matching `ListKnowledgeFolderCounts` exactly. Checked every caller
(`internal/handler/organization.go`, `internal/application/service/knowledgebase.go`,
`internal/application/service/kbshare.go`) — all of them use the result as a user-facing
"how many documents" count, so none relied on the old behavior of including deleting rows.

### Scope note

The issue also reports that housekeeping never recovers rows stranded in `deleting` status
(the second, larger symptom). That is not addressed by this PR — it's a separate behavioral
question (what should actually reclaim/finalize a row that's been stuck in `deleting` past
some threshold) and deserves its own investigation and fix rather than being bundled into a
one-line query fix.

### What was tested

- New test `TestCountKnowledgeByKnowledgeBaseID_ExcludesDeleting` in
  `internal/application/repository/knowledge_count_test.go` (in-memory SQLite, same
  `setupKnowledgeTestDB`/`insertKnowledgeInFolder` helpers `knowledge_folder_test.go` already
  uses): confirmed it fails against unmodified code (returns 3 instead of 2, counting the
  `deleting` row), then passes after the fix.
- `go test ./internal/application/repository` (the whole package, not just the new test):
  passes.
- `golangci-lint run --new-from-rev=<base> ./internal/application/repository/...` (mirroring
  this repo's own CI, which lints PRs with `only-new-issues: true` because of a large
  pre-existing backlog): 0 new issues.
- Not tested: the housekeeping/recovery behavior for stranded `deleting` rows (out of scope,
  see above).


lyingbug:
Closing as superseded by #3367, which was just merged.

This PR only covers the count-alignment slice (`CountKnowledgeByKnowledgeBaseID` excluding `deleting`). #3367 includes that same filter, plus the housekeeping Sweep C that actually recovers stranded `deleting` rows — the remaining half of #3345. Merging both would conflict on the same query.

Thanks for the focused count fix; the overlapping change is already on `main`.

PR #3377, opened 2026-09-17
fix(auth): keep login available when registration is disabled

### LINKED ISSUE #3296: [Bug]: DISABLE_REGISTRATION设为ture，已有账户登录失败
### Affected Component

Frontend UI

### Bug Description

DISABLE_REGISTRATION设为ture，已有账户登录失败。返回WeKnora 内置的 SearXNG 返回的 404 页面 HTML

### Steps to Reproduce

0. .env DISABLE_REGISTRATION设为ture
1. 填写已有账户
2. 点击登录按钮

### Expected Behavior

账户登录成功

### Actual Behavior

登录失败，返回WeKnora 内置的 SearXNG 返回的 404 页面 HTML

### WeKnora Version

v0.8.0

### Deployment Method

Docker Compose

### Operating System

Ubuntu20.04.6

### Relevant Logs

```shell

```

### Confirmation

Frank-zhu0404:
## Summary

- `DISABLE_REGISTRATION=true` / `auth.registration_mode=invite_only` must only close public `/auth/register`. Existing accounts continue to use `/auth/login` — documented on the handler and pinned with a regression test.
- Login UI: when config resolves to `invite_only`, force the login card (do not leave register mode active).
- Axios error path: do not dump HTML 404 bodies (e.g. SearXNG) into the login toast; treat them as a network/routing failure.
- `.env.example`: warn that colliding `SEARXNG_PORT` with `APP_PORT` makes `localhost:<port>` hit SearXNG first, which is the usual source of the reported HTML 404 on login.

## Test plan

- [x] `go test ./internal/handler/ -run 'TestLogin_InviteOnly|TestRegister_InviteOnly' -count=1`
- [ ] With `DISABLE_REGISTRATION=true`, `POST /api/v1/auth/login` for an existing account returns 200 JSON (not HTML)
- [ ] Login page stays on the login card under `invite_only`
- [ ] If a proxy returns HTML 404, the toast shows the network error string rather than raw HTML

Fixes #3296


lyingbug:
Thanks for the PR — closing this as it does not fix #3296.

`DISABLE_REGISTRATION` / `auth.registration_mode=invite_only` already only gates `POST /auth/register`. `POST /auth/login` never consults that flag, so the new handler comment and regression test are locking current behavior rather than changing it.

The reported failure (SearXNG's HTML 404 on login) means the browser never reached WeKnora. That is a port/proxy misroute — typically `SEARXNG_PORT` colliding with `APP_PORT` so `localhost:<port>` hits SearXNG first. Masking HTML bodies in the axios interceptor and forcing the login card in the UI does not make login succeed in that topology.

Please do not mark a follow-up as `Fixes #3296` unless login actually returns 200 JSON for an existing account in the reporter's setup (or compose/docs prevent the port collision). The `.env.example` warning is useful and can land on its own if you want to split it out.

Related investigation: https://github.com/Tencent/WeKnora/issues/3296#issuecomment-5697052831

PR #3382, opened 2026-09-18
feat(image): add a class-driven image post-processing pipeline

tayadwx:
Add a class-driven image post-processing pipeline for knowledge bases to save tokens and request count by 45+% for image processing. 

## I. Parsing Overview of the 21 Documents

- All **21 documents reached `completed`** (both parsing and summarization finished).
- **16 image-bearing documents** and 5 plain-text documents (docx / xlsx / price-list PDFs with no embedded images).
- Effective library config: `post_process_image_enabled=true`, `batch_size=4`, `classify_downscale_enabled=true`.

| Metric | Count |
|---|---|
| Total documents | 21 |
| Image-bearing documents | 16 |
| Distinct images (deduplicated) | **629** |
| chunk: text | 664 |
| chunk: image_caption | 633 |
| chunk: image_ocr | 313 |
| chunk: parent_text / summary / table_* | 77 / 21 / 2 |

### Per-document detail (img = distinct images / cap = caption chunks / ocr = OCR chunks / txt = text chunks)

| # | Document | img | cap | ocr | txt |
|---|---|---|---|---|---|
| 1 | Document 1 | 40 | 40 | 26 | 21 |
| 2 | Document 2 | 78 | 78 | 46 | 62 |
| 3 | Document 3 | 69 | 69 | 24 | 42 |
| 4 | Document 4 | 0 | 0 | 0 | 90 |
| 5 | Document 5 | 59 | 59 | 16 | 34 |
| 6 | Document 6 | 0 | 0 | 0 | 15 |
| 7 | Document 7 | 55 | 55 | 17 | 20 |
| 8 | Document 8 | 29 | 29 | 28 | 8 |
| 9 | Document 9 | 0 | 0 | 0 | 26 |
| 10 | Document 10 | 26 | 26 | 25 | 7 |
| 11 | Document 11 | 0 | 0 | 0 | 49 |
| 12 | Document 12 | 6 | 6 | 5 | 1 |
| 13 | Document 13 | 60 | 60 | 20 | 34 |
| 14 | Document 14 | 53 | 53 | 13 | 33 |
| 15 | Document 15 | 9 | 9 | 3 | 54 |
| 16 | Document 16 | 35 | 36 | 31 | 9 |
| 17 | Document 17 | 2 | 2 | 1 | 1 |
| 18 | Document 18 | 0 | 0 | 0 | 11 |
| 19 | Document 19 | 23 | 23 | 17 | 102 |
| 20 | Document 20 | 37 | 40 | 22 | 21 |
| 21 | Document 21 | 48 | 48 | 19 | 24 |

---

### 2.1 Classification distribution (629 distinct images, deduplicated by url)

Classification is produced in a single batched-description VLM call, and the result is stored in each chunk's `image_info.class` field.

| Class | Count | Share | OCR policy for this class | Triggers OCR? |
|---|---|---|---|---|
| photo | 210 | 33.4% | OCR=false | skipped |
| chart | 205 | 32.6% | OCR=true | triggered |
| decorative (non-textual / decorative image) | 88 | 14.0% | OCR=false (also `disabled`: reference removed) | skipped |
| text_screenshot (screenshot of text) | 75 | 11.9% | OCR=true | triggered |
| other | 25 | 4.0% | OCR=true | triggered |
| logo | 16 | 2.5% | OCR=false | skipped |
| table_image (table image) | 11 | 1.7% | OCR=true | triggered |
| **Total** | **629** | 100% | — | — |

### 2.2 Processing action statistics

| Action | Count | Notes |
|---|---|---|
| **OCR calls** | **316** | Only for the 4 OCR=true classes (chart/text_screenshot/table_image/other); one full-resolution image per call (`img_count=1`). 3 returned empty text → 313 ocr chunks |
| Batched description + classification calls | 169 | One request containing ≤4 images (`img_count>1`), producing both `class` and `caption` |
| Single-image supplementary description calls | 74 | Images the batch protocol failed to tag, described individually (`img_count=1`) |
| **Description / classification calls total** | **243** | 169 + 74 |
| **Total image-related VLM requests** | **559** | 316 + 243 |
| **Total tokens consumed** | **229,102** | input 170,125 / output 58,977 |

> Key point: classification and description are completed in the **same VLM call**, so classification adds zero extra overhead.

---

## III. Estimate of What Legacy Mode Would Have Cost (No Image Post-Processing)

In legacy mode (`post_process_image_enabled=false`): **every image unconditionally gets 1 caption call (full resolution) + 1 OCR call (full resolution)** — no classification, no batching, no downscaling.

| Dimension | Actual (new feature) | Estimated (legacy) | Multiple |
|---|---|---|---|
| Image VLM total requests | 559 | 629×2 = **1,258** | **2.25×** |
| └ Description / classification calls | 243 | 629 (per-image, full res) | — |
| └ OCR calls | 316 | 629 (per-image, full res, including 314 low-value images) | — |
| Total tokens | 229,102 | ≈ **419,000** (conservative lower bound) | **≈1.8×** |
| └ OCR tokens | 105,301 (316×333) | 629×333 ≈ 209,500 | — |
| └ Description tokens | 123,801 (243 calls, batched + downscaled) | 629×333 ≈ 209,500 (per-image, full res) | — |

**Estimation basis (transparent, conservative)**:
- OCR unit cost uses the measured average of a single full-resolution call: **333 tok/call** (derived from the OCR portion of the `img_count=1` group).
- Legacy description (caption) is estimated at "per-image full resolution" using the same 333 tok/call as a **conservative lower bound** — in reality legacy has no batching and no downscaling, so the unit cost of a full-resolution per-image caption would only be higher; hence 419K is a lower bound and the true gap is larger.
- Legacy mode also produces text/chunk extraction (chat.completion, asynq.chunk:extract, and other text-LLM calls). That part is identical across both modes and is excluded from this image-VLM comparison.

---

## IV. Net Savings

| Metric | Savings | Savings ratio |
|---|---|---|
| Image VLM request count | 1,258 − 559 = **699 calls** | **−56%** |
| Image VLM tokens | ≈ 419K − 229K = **≈190K** (lower bound) | **−45% and up** (actual is higher) |

---

## V. Value Assessment

1. **More than half the call volume eliminated**: For 629 images, the new feature sent only 559 image-VLM requests; legacy would need 1,258. **−56%**.

2. **Precisely skip low-value OCR**: Classification judged photo (210) / decorative (88) / logo (16) — **314 images (50%)** — as not needing OCR. Photos rarely contain extractable body text, decorative images are noise, and logos are brand marks. Legacy would OCR all 314 of them for nothing; the new feature skips them outright with **no loss of recall** (caption is still kept).

3. **Batching + downscaling further cut description cost**: Descriptions for 629 images were compressed from legacy's "629 per-image full-resolution calls" to "243 calls (batch size 4 + downscaling)". Classification itself adds zero extra calls — a single describe step yields both `class` and `caption`.

4. **Quality gains (not just cost savings)**: The `decorative` class uses the `disabled` policy, removing the image reference from the text **fed into the model**, reducing RAG noise and hallucination risk; meanwhile the caption is retained as an independently retrievable chunk, so recall is not harmed. **Precision up, recall not down**.

5. **Cost scales linearly with volume**: This library ran on a local 27B model (zero cloud cost), but once a cloud VLM is used, the saving of 699 calls × ~300 tok is a direct cut to the $ bill — and the more documents/images, the larger the saving.

6. **Operable and overridable per task**: The policy table (7 classes × OCR/Caption/Disabled) can be set at the knowledge-base level and also **flexibly overridden per task** in the upload / re-parse / online-edit confirmation dialogs (verified in the recent fixes), so it never locks the user in.

## Code Description
Add a class-driven image post-processing pipeline for knowledge bases, behind a single
knowledge-base-level switch (`post_process_image_enabled`, default `false`).

- **Switch off (default, legacy)**: behaviour is identical to the current upstream pipeline —
  every image gets a caption round and an OCR round, one request per image, no classification,
  no batching, no downscaling.
- **Switch on (classified pipeline)**: two rounds —
  1. *Describe round*: images are processed in batches (configurable 1–16, default 1) in a
     single multimodal request per batch; each image is classified into one of seven classes
     (chart / decorative / logo / photo / table_image / text_screenshot / other) and described.
     Optional downscaling to a max edge of 640 before the round (toggle, default on; OCR still
     runs on the original bytes).
  2. *OCR & post-processing*: a per-class policy table (`ocr` / `caption` / `disabled`) decides
     whether each class gets an OCR round and whether a class is retired — a retired class has
     its inline references removed from the document body and its derived caption rows
     soft-disabled (`is_enabled=false`), while the caption text itself stays in the table and
     remains searchable.

- Frontend: the KB editor gains the switch, batch size, downscale toggle and the 7-class
  policy table; the upload / re-parse dialogs inherit the KB settings and allow per-task
  overrides (`post_process_image_enabled`, `image_batch_size`,
  `image_classify_downscale_enabled`, `image_class_policies`) that are merged **per class** on
  top of the KB defaults.
- Backward compatibility: existing configurations that lack the new fields fall back to the
  defaults, and the default switch value keeps the legacy behaviour, so upgrading changes
  nothing for existing knowledge bases. The advanced rules JSON
  (`post_process_image_rules`) stays available through the API and composes with the class
  policy table.

This PR is self-contained: it does not depend on, and does not duplicate, any other open PR.

## Type of Change

- [x] ✨ New feature
- [x] 🎨 Refactor

## Related Issue

N/A (standalone feature)

## Testing

Verified on a clean checkout of this branch rebased onto `origin/main` (`c8ebb9ed`), on macOS
(arm64) + Docker Desktop. Both CI workflows were replicated locally with the same container
images and the same command sequences as `.github/workflows/app.yml`,
`.github/workflows/go-lint.yml` and `.github/workflows/frontend.yml`.

### Backend — `app.yml` sequence (101 packages, `docreader/` excluded)

| Step | Result |
|---|---|
| `go mod download` | pass |
| `scripts/check-license-bundle.sh` + `scripts/test-license-bundle.sh` | pass |
| `gofmt` over changed Go files (`cli/` excluded) | clean |
| `scripts/test-git-hooks.sh` | pass |
| `go vet` | pass |
| `go test -count=1` | 10 packages fail — **the same set as unmodified `main`** |
| `go build ./cmd/server` | pass |

The failing packages (`internal/agent/tools`, `internal/application/service`,
`internal/application/service/file`, `internal/datasource`, `.../connector/notion`,
`.../connector/yuque`, `internal/handler`, `internal/models/chat`, `internal/models/embedding`,
`internal/sandbox`, `internal/utils`) were reproduced in the same container against a pristine
`main` worktree, and their failing-test names are a superset of this branch's set. They are
environment-dependent — the sandbox resolves public hosts into `198.18.0.0/15`, which trips the
SSRF guard, plus OSS / sandbox / QueryTemplates cases that need network or cluster fixtures.

Diffing the failing test names as sets (subtests included): **failures introduced by this PR = 0**
(baseline 70, this branch 69). The single extra baseline failure is
`TestForkResetScriptPrunesLaterCommits`, which only fails when the worktree's gitdir is not
mounted (`fatal: not a git repository`) — a harness false positive, unrelated to this change.

Running the image-related subset alone gives **231 pass / 1 fail**, the one failure being
`TestUpdateKeepsSkillImageWhenEditorOmitsIt` — one of the environment-dependent baseline
failures listed above. No image-pipeline test fails.

**64 new test functions** are added across 8 `_test.go` files
(`internal/types/knowledgebase_image_policy_test.go`,
`internal/application/service/{image_multimodal_batch,image_scale,image_action,image_action_drop_reference,knowledge_process_config,knowledgebase_image_config}_test.go`,
`internal/searchutil/imageinfo_match_test.go`); all pass.

### `go-lint.yml` — the `only-new-issues` gate

Run with the same image as the workflow (`golangci/golangci-lint:v2.12.2`), reading the repo's
`.golangci.yml`:

```
golangci-lint run --new-from-rev=upstream/main \
  ./internal/application/service/... ./internal/types/... ./internal/searchutil/... ./internal/handler/...
```

→ **`0 issues`**.

For context: running the same packages on unmodified `main` *without* the diff filter reports
**213** pre-existing findings (`errcheck` 50 / `lll` 50 / `revive` 50 / `staticcheck` 43 /
`ineffassign` 9 / `unused` 7 / `gofumpt` 3 / `govet` 1), which is why the gate filters by diff.
The newly added code was brought to zero findings as part of this branch: long struct tags use
the repo's existing `//nolint:lll // one-line struct tag` convention, prompt/log lines were
wrapped without changing a byte of the resulting strings, the `image/*` blank imports carry
registration comments, and one dead helper that this branch had added but nothing called was
removed.

### Frontend — `frontend.yml` sequence

| Step | Result |
|---|---|
| `npm test` | 975 pass / 0 fail / 1 skipped (976 total) |
| `npm run type-check` (`vue-tsc --build`) | clean |
| `npm run build` | pass |
| `docker build -f frontend/Dockerfile` | pass |

The built UI image contains the new i18n keys (`policyDisabledCol`, `imagePipelineKbNote`), i.e.
the new UI really ships in the image.

### End-to-end on a self-hosted instance

`app` and `ui` were rebuilt from this branch's sources. A 7-image `.docx` (chart / parameter
table / text-screenshot / site photo / vendor logo / decorative banner / schematic placeholder)
was uploaded into knowledge bases carrying different configurations, and observed through the
pipeline logs, the `/spans` processing trace and the `chunks` table.

- **Legacy (switch off)**: 7 independent single-image tasks even with `batch_size=16` on the KB;
  `classify_max_edge=0`; zero classified-and-described calls; every per-image span reports
  `pipeline="legacy"` with no class; 7 caption rows + 5 OCR rows; all **7/7** body image
  references kept. Behavioural parity with the upstream pipeline.
- **Classified (switch on, batch 1)**: `classify_max_edge=640`; all 7 images classified; OCR
  skipped by class policy on 3 of them; 7 caption rows (2 decorative rows soft-disabled) + 3 OCR
  rows; **5/7** body references kept.
- **Batching (`batch_size=4`)**: one `Processing 4 image(s)` plus one `Processing 3 image(s)`,
  i.e. `ceil(7/4)` batches; every span carries `batched=true` with the matching `batch_size`.
- **Downscaling off**: `classify_max_edge=0` while still running the classified pipeline. The
  previous parse's rows (7 captions + 4 OCR) were soft-deleted (`deleted_at` set), so the live
  row set is exactly the new one — no leakage or double counting across re-parses.
- **Per-task overrides**: with a knowledge base deliberately left on legacy / `batch_size=1` /
  downscaling on, an upload carrying `post_process_image_enabled=true`, `image_batch_size=2`,
  `image_classify_downscale_enabled=false` and
  `image_class_policies={table_image:{ocr:false}, logo:{disabled:true}}` produced the classified
  pipeline (the override wins over the KB switch), batches of `2+2+2+1` with
  `classify_max_edge=0`, no OCR for `table_image` (span `ocr_skipped=class_policy`), and a
  soft-disabled `logo` caption. Body references dropped **7 → 5**.
- **Retiring semantics**: a retired class keeps its caption text and vector rows but is
  soft-disabled, and its inline reference is removed from the document body. Using the same
  7-image fixture against the KB default table (which retires only `decorative`) as a control,
  the rule `body references = 7 − number of retired classes` holds in every configuration above.

Two of the seven fixture images are classified slightly differently across runs (the site photo
as `photo`/`other`, the placeholder as `chart`) — ordinary VLM classification variance rather
than a pipeline defect. The policy wiring is confirmed independently by the one-to-one match
between each span's `class` and `policy` and the resulting `is_enabled` / OCR rows in the
database, in both directions (a class is OCR'd iff its policy has `ocr=true`).

The same behaviours were exercised earlier on a self-hosted deployment against a different
document set, including the `logo` class and the per-class OCR toggles.

Finally, the committed tip was rebuilt into the `app` image (`docker build` exit 0) and re-run
once end to end on the same 7-image fixture with `batch_size=3`, to make sure the shipped
artefact — not just the tree used for the cases above — behaves the same way:

- `Processing 3 image(s)` ×2 + `Processing 1 image(s)` ×1 = `ceil(7/3)`;
- all 7 spans `pipeline="classified"`, 6 of them `batched=true, batch_size=3` (the lone image of
  the last batch carries no `batched` flag, matching the documented behaviour of a batch of 1);
- `class` and `policy` match one-to-one; `decorative` and `logo` report
  `ocr_skipped=class_policy`, one `chart` reports `ocr_skipped=empty_or_invalid`;
- 7 `image_caption` rows (the `decorative` one `is_enabled=false`) and 4 `image_ocr` rows;
- body references **6/7** (`7 − 1` retired class), 0 soft-deleted leftovers.

## Checklist

- [x] Diff-scoped lint passes (`golangci-lint run --new-from-rev=<base>`, i.e. the same
      `only-new-issues` gate as `.github/workflows/go-lint.yml`) → `0 issues`
- [x] Self-reviewed the code
- [x] Added/updated tests covering the change (64 new Go test functions; 0 removed)
- [x] Updated related documentation — `docs/api/knowledge-base.md` now documents the
      `image_processing_config` fields, the built-in 7-class policy table, the per-class
      `ocr` / `caption` / `disabled` semantics, the per-upload `process_config` override keys,
      and the "omitted → unchanged / present → replaced" update semantics
- [x] Breaking changes are clearly called out in the description above — none: the switch
      defaults to the legacy behaviour and the new KB fields are additive


lyingbug (commented):
Thanks for the thorough write-up. Skipping OCR per image class is a worthwhile saving, and the per-class numbers make a good case for that part. At 5.4k lines, though, the PR bundles several independent changes, and it needs some work before we can review it properly:

1. **Please split it.** A suggested first PR: image classification + per-class OCR policy (the part your 314 skipped-OCR images come from). Batched description, classify downscaling (`image_scale.go`) and `drop_image_reference` can follow as separate PRs.
2. **Rebase.** `internal/application/service/image_multimodal.go` and `knowledge_process.go` now conflict with main, mostly because of #3445 (parse-path state fixes) and #3493 (extracted-image binding).
3. **Back up the retrieval claims or drop them.** The description says there is "no loss of recall" and that "retrieval quality improves", but the data shows request/token counts and class distribution only. Please add a before/after retrieval comparison on a fixed question set (e.g. hit rate / recall@k with the feature on vs off), or limit the claims to cost.
4. **`drop_image_reference` edits chunks after ingestion.** It goes through `UpdateDocumentChunk`, so each affected text block is re-embedded and gets a content revision, and on a revision conflict it just stands down. That's a separate behaviour change from OCR skipping. Please move it to its own PR and explain there why the decorative reference can't be left in place (the caption chunk is kept anyway).
5. **`go.mod`:** `golang.org/x/image` is in its own `require` block (line 95). Please move it into the main direct-dependency block and run `go mod tidy`.
6. **Docs:** `docs/api/knowledge-base.md` is deprecated. Please document the new KB fields in `website-docs/` instead (e.g. `website-docs/03-features/02-knowledge-base.md` / `03-document-parsing.md`).

tayadwx:
Thanks for the careful review — all six points were actionable, and working through them changed the
design rather than just the shape of the PR.

**1. Split — done.** #3635 is the first slice: image attribute observation + the OCR policy (the part
the skipped-OCR numbers come from). Batched description, downscaling and `drop_image_reference` follow
as separate PRs, in that order.

**2. Rebase — done.** #3635 is a single commit on top of current `main`; the conflicts you saw in
`image_multimodal.go` and `knowledge_process.go` (#3445 parse-path state fixes, #3493 extracted-image
binding) are resolved.

**3. Retrieval claims — dropped.** #3635 claims cost only (fewer OCR calls on images whose text is not
worth reading) and says so explicitly. If you would like a retrieval comparison — hit rate / recall@k
on a fixed question set, feature on vs off — we will run it and bring the numbers; just tell us whether
you want them in that PR or in a follow-up.

**4. `drop_image_reference` — out.** It is not in #3635. It will be its own PR, with the explicit
rationale you asked for: why the decorative body reference cannot simply be left in place when the
caption chunk is kept anyway.

**5. `go.mod` — no longer applicable.** #3635 does not touch `go.mod` at all: the `golang.org/x/image`
dependency came from the downscaling part, which is now a separate PR. That PR will put it into the main
direct-dependency block and run `go mod tidy`.

**6. Docs — done in `website-docs/`.** Fields and the decision rule in
`03-features/02-knowledge-base.md`, the two image pipelines in `03-features/03-document-parsing.md`, and
the API field summary in `04-api/02-api-knowledge.md`. The deprecated `docs/api/` is untouched.

**Where the second round took the design**

Reviewing our own first attempt, the image class turned out to be an indirection. The decision we
actually want is "does this image carry text worth a separate OCR call"; a class (chart / photo / logo)
is only a proxy for that, and it needs a new row plus a new policy entry every time a new kind of image
shows up. So #3635 replaces classification with **observation**: the describe round also answers a small
registry of attributes (`contain.text` ∈ none/sparse/block, `contain.data_visual` ∈ true/false), and a
pure function decides OCR from those.

What we think this buys:

- **The decision is code, not a model's category.** `DecideOCR` takes the observed attributes and the
  policy — no ctx, no repo, no model — so it is unit-testable and replayable.
- **Extension is additive.** A new attribute is one registry row plus, at most, one decision clause. The
  same row carries the panel's display text (label, description, meaning of each value), so the settings
  UI is generated from the registry and a new attribute needs no frontend change — which a class table
  could not offer.
- **The trace explains itself.** Each image subspan records which pipeline ran, the observed attributes,
  the decision, and why OCR was skipped, so a skipped OCR can be told apart from a failed one.
- **Failure is cheap and conservative.** An attribute the model does not answer for is left *absent*
  rather than filled with a default, and absence falls back to the conservative policy (OCR still runs).
  A prompt regression therefore costs one extra call instead of lost text, and the trace never claims an
  observation that did not happen.
- **The loop generalises.** Actions are uniform (caption / observation / ocr): the selector plans rounds,
  the dispatcher runs them, and a round's result can append another round. That is where a VLM request
  manager (retries, rate limits, timeouts) can slot in later without touching the decision logic.

**Rough plan for the follow-ups**

- **#3635**: attribute observation + OCR policy + per-image trace; off by default.
- **Batched description**: one request for N images, cutting the number of describe calls without
  changing the decision.
- **Downscaling before description**: knob in the same config; carries the `go.mod` tidy.
- **`drop_image_reference`**: the rule engine, on its own because it rewrites chunks after ingestion.
- Then, if the direction holds, the action loop plus a VLM request manager as the shared framework for
  image processing — and the question of whether a KB should be able to register attributes of its own,
  or the registry should stay release-level.

If you have a moment, directional feedback on that would be very welcome — in particular whether
attribute observation plus a code-side policy is the shape you would want for the image pipeline long
term, and whether the attribute registry should stay release-level or become something a KB can extend.
Once the slices land, would you prefer #3382 closed as superseded, or kept open as a tracker for the
sequence?

And thanks again — "split it" plus "rebase" was exactly what this needed.

PR #3407, opened 2026-09-18
feat(agent): persist compaction checkpoints across turns

lyingbug:
> **Scope note.** The token-budgeted history change (#3410) was merged into this branch and has since been **reverted** (04690a71c). A turn-by-turn simulation of `LoadAgentHistory` with the real compactor showed a flaw. The loader's budget equalled the compaction threshold, so it dropped the oldest turns before compaction could trigger. Whenever a single turn was larger than the system prompt plus the new question, nothing was ever summarized, and the session became a large sliding window with no checkpoint. That change will return as a separate PR. This PR is back to persisted checkpoints with the existing `HistoryTurns` cap, plus one flaky-test fix.

## Description

Agent context compaction was turn-local. The summary lived only in the running turn's message list, and `LoadAgentHistory` rebuilt raw history from the messages table on every turn. Once a session's history crossed the compaction threshold, **every later turn re-summarized nearly the same history**. That meant an extra LLM call close to the size of the whole window, more latency before the first token, and a history prefix that changed on every turn and so never hit the prompt cache. It is also why the `HistoryTurns` cap (default 5) can't be dropped yet even though compaction exists.

This PR persists compaction checkpoints, the same approach as Codex (`RolloutItem::Compacted`) and Pi (`appendCompaction`).

**Writing**
- A compaction whose *history* summary ends exactly on a stored turn is written onto that turn's assistant message (`messages.context_checkpoint`, JSONB). The write is column-scoped and best-effort: a failed write costs the next turn one more summarization and does not affect the current turn.
- To detect that boundary, history messages carry their turn's assistant message ID (`chat.Message.TurnID`, `json:"-"`, never sent to the provider). `compaction.Prepare` checks whether the summarized range ends on a stored-turn boundary.
- Nothing is persisted when:
  - the cut lands inside a stored turn (e.g. on a steered user message);
  - the summary covers only the live turn;
  - the history summary degraded to a raw archive. The next turn retries the summarizer instead.
- A split turn's prefix summary is left out of the checkpoint, because that turn is replayed whole next time. The `fileOps` block is recomputed from the history part only.

**Reading**
- `LoadAgentHistory` loads the newest checkpoint. The summary (as a `MessageKindCompactionSummary` message) replaces that turn and everything before it, and later turns are replayed verbatim. The compactor already treats a leading summary as `PreviousSummary`, so checkpoints fold forward incrementally.
- A checkpoint older than the fetched rows is matched by time. A failed lookup falls back to history without a checkpoint.
- The `HistoryTurns` cap still applies, now to the verbatim turns after the checkpoint. Removing the cap in favour of a token budget is a separate follow-up (see the scope note).

**Why store it on the covered turn:** the boundary is the row itself, so no request ID needs to be referenced. Session forks remap request IDs but copy the row, so the checkpoint carries over. Deleting the turn drops the checkpoint with it.

**Also in this PR:** `TestInstallSkillRecoversFromNameConflict` asserted the skill status right after `InstallSkill` returned, racing the background install, and CI hit it. It now waits for the install to finish via the existing `waitBackgroundInstallReady` helper.

**Schema:** `migrations/versioned/000104_message_context_checkpoint` and `migrations/sqlite/000024_message_context_checkpoint` each add one nullable column.

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation update
- [x] 🎨 Refactor
- [x] ⚡ Performance improvement
- [x] 🧪 Test

## Related Issue
None. #3410 was merged into this branch and then reverted; see the scope note.

## Testing
- New tests:
  - `internal/agent/compaction/checkpoint_test.go`: checkpoint on the last stored turn; stops before a split stored turn; none when the cut ends inside a stored turn, for a live-only compaction, or for a degraded summary; carries the loaded summary and its file list forward.
  - `internal/agent/engine_test.go`: `manageContextWindow` persists through the sink; a sink failure keeps the compaction; KB redaction keeps `TurnID`.
  - `internal/application/service/agent_history_checkpoint_test.go`: tagging, resume from checkpoint, checkpoint older than the window, cap after the checkpoint, lookup failure, empty checkpoint, sink wiring.
  - `internal/application/repository/message_context_checkpoint_test.go` (SQLite): newest-per-session, delete drops it, session/role scoping, and a whole-row `UpdateMessage` does not clear an existing checkpoint.
- Mutation check: removing the "same turn continues past the cut" guard fails `TestNoCheckpointWhenHistoryEndsInsideAStoredTurn`.
- `internal/database` migration tests updated (SQLite version 24, new column) and passing. SQLite up/down also run by hand with `sqlite3`. **The Postgres migration was not run against a live database.** It is a single `ADD COLUMN IF NOT EXISTS ... JSONB`, the same pattern as 000097.
- The revert was verified to leave the tree identical to the pre-#3410 head (c2a62fa5b), and the checks below were re-run on it.
- Full app package run: `go vet` over all non-docreader packages passes; `go test` over the same set passes except `TestSkillPythonVerifier`, which fails on this machine on `main` as well (local Python environment, unrelated); `go build ./cmd/server` passes.

## Checklist
- [x] Diff-scoped lint passes where applicable (for Go: `golangci-lint run --new-from-rev=origin/main ./...`)
- [x] Self-reviewed the code
- [x] Updated related documentation (`website-docs/03-features/07-agent.md`)
- [ ] Breaking changes are clearly called out in the description above (none)

## Notes for reviewers
- Existing sessions start benefiting after their next compaction; no backfill.
- Until the turn cap is removed, if more than `HistoryTurns` turns pile up after a checkpoint, the turns in between are dropped exactly as today, just with the older summary still in front.
- `AgentEngine` interface gains `SetContextCheckpointSink`; the only non-test implementation is `*agent.AgentEngine`.
- Follow-ups from review, left for the next PR:
  - Match a checkpoint that is older than the fetched rows by `(created_at, id)` rather than comparing a turn's first user time with the checkpoint's assistant time.
  - Write the checkpoint even when a compaction frees under 5% and the engine discards it.
  - Check `RowsAffected` so that a write matching no row is not logged as saved.
  - Add a partial index for `GetLatestContextCheckpoint`; `messages` only has an index on `session_id`.


internal/application/service/agent_history.go
```diff
@@ -97,30 +182,132 @@ func LoadAgentHistory(
...
-		out = append(out, buildTurnBodyMessages(t.assistant, t.users[1:])...)
+// messages replays one turn, tagged with its assistant message ID.
+func (r *historyReplay) messages(t *agentHistoryTurn) []chat.Message {
+	id := t.assistant.ID
+	if msgs, ok := r.built[id]; ok {
+		return msgs
 	}
-	return out, nil
+	msgs := append([]chat.Message{buildUserHistoryMessage(t.users[0])},
+		buildTurnBodyMessages(t.assistant, t.users[1:])...)
```
lyingbug: 这里计算应该不准确

PR #3415, opened 2026-09-18
test(im): give each lifecycle test its own sqlite database

### LINKED ISSUE #3198: [Bug]: IM lifecycle tests share one in-memory SQLite database when two tests start in the same clock tick
### Affected Component

Backend Service & API

### Bug Description

`internal/im` lifecycle tests can fail nondeterministically with

```
service_lifecycle_test.go:383: create im_channels: table im_channels already exists
```

The cause is in the test helper `newLifecycleTestDB` (`internal/im/service_lifecycle_test.go:46`):

```go
func newLifecycleTestDB(t *testing.T) *gorm.DB {
	t.Helper()
	db, err := gorm.Open(sqlite.Open(fmt.Sprintf("file:im-lifecycle-%d?mode=memory&cache=shared", time.Now().UnixNano())), &gorm.Config{})
	...
	if err := db.Exec(`CREATE TABLE im_channels (...)`).Error; err != nil {
		t.Fatalf("create im_channels: %v", err)
	}
```

The in-memory SQLite database name comes from `time.Now().UnixNano()` and nothing else. Two tests whose helper calls land in the same wall-clock tick get the *same* DSN, and a `cache=shared` in-memory database stays alive while any connection to it is open, so the second test's `CREATE TABLE im_channels` collides with the first test's table. Because the name is not unique per test, the tests also share table state instead of each getting a fresh database.

This is the only such helper in the repository: every other test database in the tree derives its name from `t.Name()` or a UUID (for example `internal/application/repository/tenant_api_key_test.go`, `internal/application/service/memory/service_test.go`, `internal/application/repository/chunk_revision_test.go`). This helper is the single outlier, so aligning it with the existing convention fixes it.

### Steps to Reproduce

1. Check out `7a98a8e3de5a9aa85c8f1fa6b02772986cb435d0` (current `main`).
2. Run the package suite repeatedly, e.g. `for i in $(seq 8); do go test ./internal/im/ -count=1; done`.
3. Observe that some runs fail with `create im_channels: table im_channels already exists`, with a different test failing each time.

Observed here on go1.26.3 / Windows 10, `CGO_ENABLED=1`, 8 consecutive runs: **6 failed, 2 passed**. Failing tests included `TestStartChannelDoesNotLeakAdapterWhenStoppedDuringFactory`, `TestEnsureChannelAdapterKeepsRuntimeOnDatabaseFailure`, `TestLeaderElectionFailsClosedWhenRedisUnavailable`, `TestLeadershipLossStopsAdapterAndSchedulesRecovery` and `TestChannelConfigEventReloadsOtherReplica` — i.e. whichever lifecycle tests happen to run back-to-back within one clock tick.

The window is the platform's wall-clock granularity: on Windows the wall clock updates in ~15.6 ms steps, so two tests that each open a database within one step collide. The same applies to any sandboxed or virtualized host with a coarse or paused wall clock. It is not related to the code under test — the affected tests are unrelated to each other and pass in isolation.

### Expected Behavior

Each test gets its own in-memory database, so the suite is deterministic and independent of wall-clock resolution, and no test can observe tables or rows created by another test.

### Actual Behavior

Database names collide whenever two helper calls fall in the same clock tick. The colliding test fails with `table im_channels already exists`, and the shared `cache=shared` database means tests can also see each other's data. Different tests fail on different runs, so a real regression in one of them can be masked by an unrelated failure.

### WeKnora Version

commit `7a98a8e3de5a9aa85c8f1fa6b02772986cb435d0` (built from source)

### Deployment Method

Build from source

### Operating System

Windows 10 Pro (any host with coarse wall-clock granularity is affected)

### Relevant Logs

```shell
--- FAIL: TestStartChannelDoesNotLeakAdapterWhenStoppedDuringFactory (0.00s)
    service_lifecycle_test.go:383: create im_channels: table im_channels already exists
--- FAIL: TestLeadershipLossStopsAdapterAndSchedulesRecovery (0.00s)
    service_lifecycle_test.go:383: create im_channels: table im_channels already exists
FAIL
FAIL	github.com/Tencent/WeKnora/internal/im	0.467s
```

### Confirmation

ankit373:
## Description

`newLifecycleTestDB` (`internal/im/service_lifecycle_test.go`) derived its in-memory SQLite DSN from `time.Now().UnixNano()` alone. Two tests whose helper calls land in the same wall-clock tick get the same DSN, and since the database is `cache=shared`, the second test's `CREATE TABLE` collides with the first's:

```
service_lifecycle_test.go:383: create im_channels: table im_channels already exists
```

A shared database also means one test can observe rows a different test created.

## Fix

Every other in-memory test database in this repo already names itself from `t.Name()` instead of the wall clock (`internal/application/repository/tenant_api_key_test.go`, `internal/application/service/memory/service_test.go`, and others); this helper was the only outlier. Switched it to the same convention. `t.Name()` is unique per test function, which is all this needs since none of these lifecycle tests call the helper from subtests.

## Testing

Added `TestNewLifecycleTestDBIsolatesConcurrentTests`, which runs the seed and the isolation check as actual parallel subtests (`t.Parallel()`) rather than a sequential pair, closer to the real collision scenario.

Confirmed the test fails without the fix: forcing the DSN back to a fixed name reproduces `table im_channels already exists` immediately. With the fix, ran the full package as 8 separate `go test ./internal/im/ -race` invocations (separate processes, matching how the original race actually surfaces) with no failures.

`golangci-lint run --new-from-rev` couldn't run in my environment (local binary was built with go1.25, this repo targets go1.26.0), so I'm recording that here per the template instead of a clean run. `gofmt` and `git diff --check` are both clean.

## Related Issue

Fixes #3198

## Note for reviewers

#3199 diagnosed and fixed this same issue with the same `t.Name()` approach, but was closed without comment and the fix wasn't picked up elsewhere. I independently reached the same core fix, plus a parallel-subtest regression test instead of a sequential one.


lyingbug:
Thanks for working on the lifecycle test isolation. This version introduces a reproducible regression when tests are repeated in one process.

At head 2c1408f3ac473c3ce6e1427d6f83aafbd4000d0f, internal/im/service_lifecycle_test.go:49 uses t.Name() as the shared in-memory SQLite database name, but newLifecycleTestDB never closes the underlying sql.DB. t.Name() repeats across -count iterations, so the second iteration reconnects to the still-live database and CREATE TABLE fails.

Reproduced locally with:

```sh
go test ./internal/im -run '^TestNewLifecycleTestDBIsolatesConcurrentTests$' -count=2 -timeout=90s
```

The second iteration fails with `table im_channels already exists` in both subtests. Separate go test processes do not exercise this case.

Please register sql.DB.Close with t.Cleanup immediately after opening the database, or give each helper invocation a unique database identity and still clean up its connections. Keep a repeated-run check alongside the concurrent isolation test.

Closing the current version because it regresses repeatable test execution; a corrected follow-up is welcome.

ankit373:
You are right, and thanks for the exact reproduction. Reproduced it on `2c1408f3` before touching anything: both subtests fail with `table im_channels already exists` on the second iteration.

The follow-up is #3433. It does both things you suggested rather than picking one, because closing alone turned out not to be sufficient: a single test that calls the helper twice holds both databases open before any cleanup runs, so the name still has to be unique per invocation. `t.Cleanup` now closes the `sql.DB` immediately after `gorm.Open` (registered before the caller's own cleanups, so `t.Cleanup(svc.Stop)` still runs first), and the name carries a process-wide sequence number.

Both the repeated-run check and a connection-closed check are in the suite, and I mutation-checked each half so neither is decorative: dropping the cleanup fails one, dropping the sequence fails the other.

On how this got past me: I verified #3415 by running the package as several separate `go test` processes rather than one process with `-count>1`, which as you noted is exactly the case that cannot exercise this. That was the wrong check and the regression was mine.

PR #3443, opened 2026-09-20
fix(wiki): renew and owner-scope the per-slug reduce lock

Fuzzycorrect:
## Description

The per-slug lock that serializes Reduce on a shared wiki page is taken with `SetNX(key, "1", 5m)` and released with an unconditional `defer Del(key)`. Three problems follow, all from the same root cause: the hold-and-release semantics are wrong.

1. **The lock expires while its owner is still writing.** `wikiSlugLockTTL` is 5 minutes, but one reduce is a single LLM page-modify call over a large page — measured on an 87-document KB at **p50 5.3m / p90 17.2m**. A second batch therefore acquires the lock and starts its own read-modify-write on the same page. That is the concurrent-writer window behind the slug duplicate-key failures and the silently lost contributions.
2. **Release ignores ownership.** Because the TTL had usually already lapsed, the original holder's bare `DEL` deletes a key that by then belongs to a *different* batch — handing the page to a third writer in the middle of the second one's write.
3. **The wait bound is far too short.** `wikiSlugLockWait` is 2 minutes, i.e. **2.5× shorter than the mean page write**. A reducer that politely waits is told "slug busy" and abandons the update; the caller then re-queues **every contributing document** and re-runs extract + classify + summary + all of its pages.

### Measured on an 87-document KB (before)

| Measurement | Value |
| --- | --- |
| Same-slug page spans with overlapping intervals | 129 pairs across 73 pages (a correct mutex gives 0) |
| Page writes exceeding the 2m wait bound | 1,232 / 3,927 (31%) |
| Page writes exceeding the 5m TTL | 926 / 3,927 (24%) |
| Page-write cost that was pure redo | 1,370 / 3,825 (36%) |
| Amplification of one blocked page | a 2-minute block → **~111-minute** full document re-ingest |

### Measured after this change (same KB, same load)

| Signal | Before | After |
| --- | --- | --- |
| Same-slug overlapping page spans (post-fix window) | 129 cumulative | **0** |
| `deferring update` (30-minute window) | 3 / hour | **0** |
| slug duplicate-key errors | 7 / 20h | **0** |
| Live lock values | all the constant `"1"` | **10 distinct UUIDs** |
| Lock TTL while held | monotonically falls to 0 | **renewed back to ~300s** |
| Second batch blocked on a contended slug | abandoned at 2m | **waited 3m21s, then acquired and completed** |

## Type of Change

- [x] 🐛 Bug fix
- [x] 🧪 Test

## Related Issue

No linked issue.

## Testing

Run against this branch on top of `main` (`68e21b6`):

- `go build ./internal/application/service/` — clean
- `go test ./internal/application/service -count=1` — pass
- `go test ./internal/application/repository -count=1` — pass (unaffected package, sanity check)
- `gofmt -l` on the changed files — clean
- `git diff --check HEAD~1...HEAD` — clean

New tests in `internal/application/service/wiki_slug_lock_test.go` (miniredis):

- `TestWikiSlugLockScriptsRequireOwnership` — a non-owner's renew and release are both no-ops and leave the key intact; the owner's are not. This is the property that stops a TTL-overrunning holder from deleting someone else's lock.
- `TestWithSlugLockSerializesSameSlug` — three concurrently contending reducers never interleave inside the critical section. This is the property the lock exists for.
- `TestWithSlugLockReleasesOwnLock` — the key is present while `fn` runs and gone once it returns.

The tests deliberately do not depend on wall-clock time (`miniredis.FastForward` does not advance a real `time.Ticker`), so they assert the Lua scripts' ownership semantics directly.

Note on scope: `go build ./...` fails in my environment on `github.com/asg017/sqlite-vec-go-bindings/cgo` because `sqlite3.h` is not installed. That is a pre-existing environment gap unrelated to this change, which is why the build/test commands above are scoped to the affected packages.

## Checklist

- [ ] Diff-scoped lint passes where applicable (for Go: `golangci-lint run --new-from-rev=origin/main ./...`) — **not run**, `golangci-lint` is not installed in my environment
- [x] Self-reviewed the code
- [ ] Updated related documentation (README, `docs/`, Swagger annotations, etc.) — N/A, no user-facing or API surface change

## Screenshots / Recordings

N/A — no UI change.


lyingbug (commented):
Thanks for the careful measurements. Owner-scoped release plus a heartbeat is the right shape for this lock, and we'd like to merge it. Two of the timing constants need another look first:

1. `wikiSlugLockRenewMax = 30 * time.Minute` is shorter than the task budget: wiki ingest tasks run with `asynq.Timeout(60*time.Minute)` (`wiki_ingest_batch.go`, `wiki_ingest.go`). A reduce that is still running at ~30m stops renewing, and the key expires about `wikiSlugLockTTL` later. Past roughly 35m, a second batch can take the slug while the first is still writing, which is the same overlap this PR removes for shorter reduces. The holder's `ctx` is already bounded by the task timeout, so renewing until `renewStop` or `ctx.Done()` already stops a wedged holder. I'd suggest dropping the separate cap. If you want to keep it, please set it to at least the task timeout.
2. `wikiSlugLockWait = 20 * time.Minute`: while a goroutine polls for a contended slug, it holds one of the `reduceParallel` slots of `egReduce` (`wiki_ingest_batch.go`, around line 580). A few hot slugs can therefore stall the rest of the batch's reduces for up to 20m each, all inside the same 60m task budget. Please either derive the wait from the time remaining on `ctx` (e.g. leave headroom for the reduce itself), or explain in the comment why 20m is safe when this interacts with `reduceParallel` and the task timeout.

Optional / non-blocking: a short test showing that renewal stops on `ctx.Done()` would be useful once the cap logic changes.

PR #3444, opened 2026-09-20
fix(queue): order ClaimBatch by fail_count so retries cannot starve new work

Fuzzycorrect:
## Description

`ClaimBatch` picks work with `ORDER BY id ASC` (Postgres) / `Order("MIN(id) ASC")` (GORM), i.e. oldest-first by primary key.

Retry does not move a row. `requeueFailedOps` calls `IncrFailCount` and then `ReleaseByIDs`, which deliberately keeps the row in place so the `fail_count` budget keeps counting down.

Together those two produce a starvation mode: a document that fails repeatedly keeps its **original, lowest id**, so it is always selected first — while every never-attempted document, enqueued later and therefore holding a higher id, waits behind it until the stuck op finally exhausts `wikiMaxFailRetries` and dead-letters. A failure mode that blocks a handful of documents (e.g. per-slug lock contention) therefore turns the backlog into a standstill rather than a slowdown.

### Observed on a real 87-document KB

```
   op   | 总数 | 已认领 | 最小id | 最大id
--------+------+--------+--------+--------
 ingest |   42 |      4 |   7271 |   7317
```

At the same moment asynq reported `pending=0 scheduled=0 retry=0 archived=0 active=4`. With 8 configured workers, `active=4` and an empty pending queue mean the bottleneck was **not** throughput:

- the four ops at the head of the queue were **all re-runs of documents that already had a root page**;
- **38 never-attempted documents** (strictly higher ids) were never claimed;
- over 20 hours the KB completed only 47 of 87 documents.

## Type of Change

- [x] 🐛 Bug fix
- [x] 🧪 Test

## Related Issue

No linked issue.

## Testing

Run against this branch on top of `main` (`68e21b6`):

- `go test ./internal/application/repository -count=1` — pass (whole package, including the pre-existing FIFO tests)
- `gofmt -l` on the changed files — clean
- `git diff --check HEAD~1...HEAD` — clean

New test: `TestTaskPendingOps_ClaimBatch_PrefersLeastFailed` replays the real retry path on SQLite — claim → `IncrFailCount` → `ReleaseByIDs` — then asserts that the never-attempted document (higher id, no failures) is claimed ahead of the retried one (lower id, one failure), and that the retried document is still eventually picked up rather than stranded.

**Backward compatibility:** when every row has `fail_count = 0` the new ordering degenerates exactly to the previous `id` ordering, so the existing `ClaimBatch` FIFO tests pass unchanged. Only rows that have already failed are demoted — which is the intent.

Both branches are covered: the Postgres anchor query orders by `fail_count ASC, id ASC` at both the window-function level (so the anchor row is the least-failed one) and the anchor-to-key mapping, and the GORM branch uses `Order("MIN(fail_count) ASC, MIN(id) ASC")`.

## Checklist

- [ ] Diff-scoped lint passes where applicable (for Go: `golangci-lint run --new-from-rev=origin/main ./...`) — **not run**, `golangci-lint` is not installed in my environment
- [x] Self-reviewed the code
- [ ] Updated related documentation (README, `docs/`, Swagger annotations, etc.) — N/A, no user-facing or API surface change

## Screenshots / Recordings

N/A — no UI change.


ben0112:
Confirming this mechanism at a much larger scale — we hit the same starvation on a **24,624-op** wiki backlog (v0.8.0, single KB, bulk import).

**Symptom** — the same shape as yours:

```
asynq:{wiki}   pending=0  scheduled=0  retry=0  archived=0  active=4
task_pending_ops  (wiki:ingest)  24,624 rows, 23,089 never claimed
```

**Why** — the already-failed rows own the lowest ids, so `ORDER BY id ASC` pins them at the head permanently:

```
min(id) all rows            = 55227
min(id) rows fail_count > 0 = 55227   <-- failed rows own the head
max(id) rows fail_count > 0 = 57631
min(id) rows fail_count = 0 = 56178   <-- every never-attempted row sits behind them
max(id) all rows            = 84604
```

The first 8 rows a batch claims, straight from `ORDER BY id ASC`:

| id | fail_count | claimed |
|---|---|---|
| 55227 | 4 | t |
| 55374 | 2 | t |
| 55862 | 2 | t |
| 55897 | 1 | t |
| 55904 | 1 | t |
| 56000 | 1 | t |
| 56002 | 1 | t |
| 56178 | 0 | t |

Every claim round re-selects the same failed head — they retry, `IncrFailCount` + `ReleaseByIDs` keeps their id, so ids 56178–84604 were never reached once. Over ~17h the KB completed ~200 of its 24,300 pending documents while `active` sat at 3–4 with an empty queue, and `COUNT(*) WHERE claimed_at IS NULL` stayed frozen at 23,089 for hours.

**Confirmation of the mechanism**: deleting the 17 `fail_count > 0` rows moved the head to 56178 and the queue immediately began claiming never-attempted work. That is a destructive workaround, not a fix — it just demonstrates the ordering is what starves the backlog.

Your `fail_count ASC, id ASC` ordering is exactly what this needs. The degenerate case you call out (`fail_count = 0` everywhere → identical to the old `id` ordering) is also what makes the interim workaround safe.

**One adjacent observation** that may matter for the same KBs: the starved documents also stay `parse_status = finalizing`, and housekeeping deliberately skips them — it reports `N candidate(s) skipped — tasks still queued (backpressure, not stuck)`. In our case it skipped 24,041 of 24,299 candidates per sweep and recovered 3 rows in 17 hours. So a starved `ClaimBatch` presents to the operator as **permanently-processing documents with an idle model server and nothing in the logs**, which is a fairly opaque failure mode — it took us a while to get from "GPU utilisation dropped to 0" to this ordering. If the queue-depth view could distinguish "claims are failing repeatedly at the head" from "waiting on workers", that alone would have shortened the hunt.

lyingbug:
Review follow-up pushed to this branch (`b5ee469f9`).

The original `ClaimBatch` change is the right fix for the standard-mode starvation: a retried row keeps its id, so `ORDER BY id ASC` pins failed docs at the head and never-attempted work never gets claimed.

This branch still left the Lite path on `PeekBatch` (`ORDER BY id ASC`). Default ingest batch size is 5, so the same head-of-line stall happens there when failed docs occupy the first page. The follow-up commit:

- orders `PeekBatch` by `fail_count ASC, id ASC` (same rule; all-zero `fail_count` is still FIFO)
- adds `TestTaskPendingOps_PeekBatch_PrefersLeastFailed`
- updates the stale comments that still described `id ASC` / “earliest eligible id”

Two leftovers that do not need to block merge:

1. The new tests run on SQLite (GORM `MIN(fail_count), MIN(id)`). Production standard mode uses the Postgres window + outer `ORDER BY fail_count, id`. Those are equivalent for the starvation case (never-failed keys beat failed keys), but they can pick a different key when one `dedup_key` mixes fail counts (GORM tie-breaks on the group’s `MIN(id)`, Postgres on the least-failed row’s id).
2. Least-failed-first means a continuous stream of new uploads can delay retries until the `fail_count = 0` queue drains. Fine for a finite backlog; worth knowing if a KB is fed continuously.

`go test ./internal/application/repository -run 'TestTaskPendingOps_PeekBatch|TestTaskPendingOps_ClaimBatch'` passed on the follow-up.

PR #3476, opened 2026-09-20
feat(datasource/yuque): preserve TOC folder hierarchy for synced documents

RichardChengshaojin:
## Description

Documents synced through the Yuque datasource connector all land flat at the knowledge base root — the book's TOC hierarchy is discarded.

`folder_path` is derived from exactly one input: the `FileName` the connector emits (`SplitKnowledgeRelativePath` → `NormalizeKnowledgeFolderPath`). The Yuque connector currently emits a bare `<title>.md`, so the derived folder is always empty and every document ends up at the root.

This is **not a new pattern for this codebase**: the GitLab connector already emits a path-qualified `FileName` (`project-branch/sub/dir/file.md`) and therefore produces a proper hierarchy. This PR brings the Yuque connector in line with that — it supplies the missing input, and the existing `FileName → folder_path` machinery does the rest.

**Changes**

| File | Change |
|---|---|
| `yuque/client.go` | `ListBookTOC()` — one request per book per sync (the endpoint returns the whole tree, unpaginated) |
| `yuque/types.go` | TOC response types, `buildTOCPaths()` (pure), folder-path settings, `buildFolderFileName()` |
| `yuque/connector.go` | `walk()` fetches the TOC per book, folds it into a per-document path, and optionally filters to TOC-listed documents |
| `yuque/toc_test.go` | New — 23 test functions |

The book title used for the leading path segment comes from `detail.Book.Name`, which is already present in the per-document detail response, so no additional API call is introduced.

```go
// per document
fileName := buildFolderFileName(settings, tocOK, bookName, tocSegs[d.ID], d.Title)
```

Path segments go through the shared `datasource.SanitizeFileName` (recently extracted from the per-connector copies), so a `/` inside a group title cannot create an unintended nesting level.

### Configuration

Three optional keys under `DataSourceConfig.Settings`, read the same way the GitLab connector reads `Settings["projects"]`:

```json
{ "folder_mode": "toc", "include_book_title": true, "toc_only": false }
```

| Key | Values | Default | Effect |
|---|---|---|---|
| `folder_mode` | `none` / `toc` | **`none`** | `none` = current behaviour; `toc` = derive the folder from the TOC |
| `include_book_title` | bool | **`true`** | Prefix the path with the book title |
| `toc_only` | bool | **`false`** | Skip documents that do not appear in the TOC at all |

### Backward compatibility

**`folder_mode` defaults to `none`, so upgrading changes nothing.** The emitted `FileName` stays byte-for-byte identical to today's output, and a dedicated unit test asserts exactly that. Any unrecognised or malformed settings value falls back to the default rather than failing the sync.

### Error handling

- TOC request fails or returns malformed data → log a warning and fall back to flat names. **The sync must never fail because of this.**
- A document present in the book but absent from the TOC → no folder segment; it stays at the KB root.
- `parent_uuid` pointing at a missing node, or a cycle in the TOC → the upward walk terminates via a visited set (plus a depth bound); no infinite loop.
- Book title empty → fall back to the namespace; if that is empty too, emit no prefix and do not error.

### `toc_only` semantics — please note the distinction

The predicate is **"does this document appear in the TOC node list"**, *not* "does it have a non-empty path". The two are not equivalent: a document can legitimately appear in the TOC with **no enclosing group**, in which case `inTOC == true` and its path is empty.

| Document | `inTOC` | path | kept with `toc_only=true` |
|---|---|---|---|
| filed under a group | true | `Group/Sub` | yes |
| **in the TOC, but no group** | **true** | **`""`** | **yes** — must not be filtered |
| present in the book, absent from the TOC | false | — | no |

Using "empty path" as the filter predicate would silently drop the middle row, which is why the implementation keeps **two separate maps**. There is a regression test for it.

Relatedly, the filter is applied **only when the TOC was actually fetched** (a `tocOK` guard). Without that guard a failed TOC request would leave `inTOC` empty and `toc_only=true` would drop the entire book — an outage would silently become data loss. When the TOC is unavailable the connector degrades to the previous flat behaviour instead.

### Behaviour change to be aware of

Once `folder_mode=toc` is enabled, the folder layout is owned by the connector and re-derived on every rebuild, so manual folder edits in the UI will be overwritten by the next sync. That is why the mode is opt-in and defaults to off.

## Type of Change

- [x] ✨ New feature

## Related Issue

No existing issue — happy to open one for tracking if that is preferred.

## Testing

**Unit tests** (`go test ./internal/datasource/connector/yuque/`)

23 test functions covering:

- `buildTOCPaths`: empty / title-only TOC, single and multi-level nesting, DOC nested under DOC, dangling `parent_uuid`, **cycle termination**, per-segment sanitisation of titles containing `/`, non-DOC nodes, empty `doc_id`, duplicate `doc_id`, and the `inTOC` / `path` separation
- `flexibleDocID`: the TOC `doc_id` field is a **number on DOC nodes but an empty string on TITLE nodes**, so plain `int64` decoding fails on real payloads — covered by a realistic mixed-shape payload test
- `buildFolderFileName`: **byte-identical output under default settings** (regression guard), with/without the book segment, empty book title, book title needing sanitisation, and degradation when the TOC is unavailable
- `parseFolderSettings`: defaults, all values read, and tolerance of unknown/wrongly-typed values
- End-to-end through `FetchAll` against an `httptest` server: path assembly, `toc_only` keeping a group-less document, **TOC 404 degrading without error and without filtering**, and two books sharing a top-level group name staying separate

**Also run**

```
gofmt -l               # clean
go vet                 # clean
go build ./...         # BUILD OK
golangci-lint v2.12.2 run --new-from-rev=HEAD ./internal/datasource/connector/yuque/...   # 0 issues
git diff --check       # clean
```

**Manual verification against a real Yuque workspace**

Enabled `folder_mode=toc` on a datasource syncing one book (6 documents across 4 top-level groups) and compared the resulting folder tree with the workspace's actual TOC. All 6 documents landed exactly where the TOC places them, including documents filed directly under the book rather than under a group. This also exercised `SanitizeFileName` on a real title containing `|` — it was replaced with `_` without splitting the path into an extra level.

## Checklist

- [x] Diff-scoped lint passes (`golangci-lint run --new-from-rev=origin/main ./...` — 0 issues)
- [x] Full-repository checks were run (`go build ./...` passes; unit tests scoped to the changed package pass)
- [x] Self-reviewed the code
- [ ] Updated related documentation — **not applicable**: there is no datasource/connector settings document under `docs/`; the new settings are documented in the code. Frontend controls for the two user-facing switches are intentionally a follow-up PR so this one stays reviewable.
- [x] Breaking changes are clearly called out in the description above — none; the feature is opt-in and the default output is unchanged

## Screenshots / Recordings

Not applicable — no UI change in this PR.


lyingbug (commented):
Thanks — the TOC handling is careful (the `inTOC` vs. path distinction and the `tocOK` guard are exactly right), and we'd like to merge this. Two requests:

1. **Make the feature reachable from the UI in this PR.** `folder_mode` defaults to `none` and `DataSourceEditorDialog.vue` has no control for it, so after merging, the only way to use it is hand-crafting `config.settings` via the API — effectively nobody will. The dialog already keeps unknown `settings` keys, so a small Yuque-only select for `folder_mode` (plus the `include_book_title` / `toc_only` switches, or at least `folder_mode`) with i18n strings would be enough. We'd rather ship backend and toggle together than track a follow-up.
   Related question: since the "folder layout is owned by the connector" caveat mainly affects existing sources, would you be open to defaulting `folder_mode` to `toc` for **newly created** Yuque sources (set by the frontend on create), while existing sources keep `none`? Most users expect the book's TOC hierarchy.
2. **Document the settings.** There is a Yuque section in `website-docs/03-features/10-datasource.md` (`#### Yuque 语雀`) — please add the three `settings` keys, their defaults, and the "manual folder edits get overwritten on sync when `toc` is on" caveat there (the old `docs/` directory is deprecated, which is probably why you didn't find a place).

RichardChengshaojin:
Thanks — both requests are in the latest push.

**UI.** `folder_mode` gets a Yuque-only control in the sync-strategy step. I used the pill-button group that the neighbouring sync-mode and conflict-strategy settings already use rather than introducing a select, so it reads as one of them. `toc_only` sits below it, disabled unless the TOC layout is selected — the connector only fetches the TOC under `folder_mode=toc`, so the flag is inert in the flat layout and the control says so.

On your question: **yes, defaulting `folder_mode` to `toc` for newly created sources is fine by us.** The frontend writes the key at creation only, and editing an existing source never injects it, so the "folder layout is owned by the connector" caveat cannot reach a source that predates the change. The backend default stays `none`, so the byte-for-byte regression guard is untouched.

**Docs.** Added to the Yuque section of `website-docs/03-features/10-datasource.md`, including the overwrite caveat — thanks for the pointer to that file.

**Two changes beyond the review, both worth a look while you're here:**

- **`toc_only` is now an admission filter, not a reaper.** It used to be applied before the document was recorded for deletion detection, so a document present in the source but not admitted came back as `IsDeleted` — and with `sync_deletions` on, the sync service deleted it from the knowledge base. On Yuque that is especially costly: the web UI does not show documents created through the API that were never attached to the TOC, so a knowledge base is often the only place they can be reached at all. A filtered document is now simply not ingested; an already-ingested copy is left alone, and it stays out of the cursor so that admitting it later (by attaching it to the TOC, or by switching the filter off) ingests it rather than dismissing it as unchanged.
- **`include_book_title` is gone.** Leaving the book segment out silently merges same-named top-level groups from different books, and the path is the only carrier of that dimension — the automatic tag granularity is one tag per data source, not one per book. The book segment is now unconditional under `folder_mode=toc`.

A screenshot of the new controls is in the description, along with the full rationale, the tests and the settings table.

weknora-yuque

PR #3477, opened 2026-09-20
fix(session): reject invalid QA queries before session lookup

### LINKED ISSUE #2057: [Feature]: KnowledgeQA 入口尽早校验 query，避免跑完检索才失败
### Affected Component

Backend Service & API

### Problem Description

ValidateInput（控制字符 / UTF-8 / XSS）只在 INTO_CHAT_MESSAGE 拼 prompt 前执行；HTTP 入口对 query 仅有 binding:"required"。恶意或非法输入会先跑完检索/重排/Merge，再失败。

### Proposed Solution

建议：
Handler/Service 入口 fail-fast（KnowledgeQA 等）
拼 prompt 前保留校验（纵深防御，覆盖 Agent 等非 HTTP 入口）

收益： 
省成本、失败更早更清晰（如 400）、安全边界更靠前。

### Alternatives

_No response_

### Impact

Many users would benefit

### Use Case

_No response_

### Additional Information

_No response_

### Confirmation

yushangcan:
## Description
QA requests containing control characters or content rejected by the existing `ValidateInput` rules currently pass HTTP request parsing and can start session lookup and retrieval before failing during prompt construction. Validate queries immediately after JSON binding in the shared parser for `KnowledgeQA` and `AgentQA`, returning HTTP 400 for invalid or whitespace-only input.

Valid queries retain their original text. The existing prompt-construction validation remains in place. Regression tests exercise both HTTP handlers and verify that rejected input never reaches session lookup.

## Type of Change
- [x] Bug fix
- [x] Test

## Related Issue
Fixes #2057

## Testing
Validated the unchanged source and test files with Go 1.26.8 on Linux, with CGO enabled:
- Before the fix, 14 regression cases failed because invalid queries reached session lookup.
- `go test ./internal/handler/session -count=1`: passed, including all 24 new regression cases across both endpoints.
- `go vet ./internal/handler/session`: passed.
- `golangci-lint run --new-from-patch=/tmp/qa-query-validation.patch --timeout=8m ./internal/handler/session/...`: passed with 0 issues. The patch included both the implementation and the new test file.
- `gofmt -l` on both changed files and `git diff --check origin/main...HEAD`: passed.
- Verified a clean merge with current `origin/main` using `git merge-tree --write-tree origin/main HEAD`.

Validation was scoped to the changed package; full-repository tests and live model calls were not run.

## Checklist
- [ ] Full-repository checks were run
- [x] Self-reviewed the code

## Screenshots / Recordings
Not applicable (backend-only change).


lyingbug (commented):
Thanks. Rejecting bad input before session lookup is the right move, and we'd like to merge this after one scope fix:

1. `parseQARequest` in `internal/handler/session/qa.go` is shared by `KnowledgeQA` and `AgentQA`, so `AgentQA` now goes through the full `secutils.ValidateInput`, including its `xssPatterns` (`internal/utils/security.go`, lines ~29-46 and ~98-103). Until now the agent path never ran that check: only the KnowledgeQA pipeline does, in `chat_pipeline/into_chat_message.go`. After this change, an agent question like "why doesn't my `<button onclick=...>` fire?" or "what does `javascript:void(0)` do?" gets a 400. That's a regression for a common use case, since users paste frontend code into agent chats. Please narrow the entry check to what is safe for both endpoints: control characters, UTF-8 validity, and whitespace-only input. Leave the XSS-pattern check where it is today in the KnowledgeQA pipeline. Alternatively, run the full `ValidateInput` only when `logPrefix`/the endpoint is KnowledgeQA. It would help to add an `AgentQA` test case with an HTML/JS snippet that is expected to succeed.

PR #3492, opened 2026-09-20
fix(wiki): reject length-truncated page rewrites (fixes #3468)

### LINKED ISSUE #3468: [Bug]: wiki page rewrites are written back unvalidated — a truncated LLM rewrite silently shrinks the page (156 pages / 88,556 chars lost in our KB)
### Affected Component

Frontend UI

### Bug Description

Every wiki page update hands the whole page to the chat model and takes back a full rewrite (wiki_ingest_batch.go:2042, agent.WikiPageModifyUserPrompt), then writes the result straight back:

  wiki_ingest_batch.go:2065  updatedSummary, updatedBody := splitSummaryLine(updatedContent)
  wiki_ingest_batch.go:2067  page.Content = updatedBody
  wiki_ingest_batch.go:2069  page.Content = updatedContent
  wiki_ingest_batch.go:1843  page.Content = summaryUpdate.SummaryBody   (summary pages, same pattern)

Nothing between the call and the write compares input with output. Two ways this loses data silently:

1. The model stops early. generateWithTemplate sets MaxTokens: wikiLLMMaxTokens (32768, wiki_ingest.go:2553) but returns only the text — return response.Content, nil (wiki_ingest.go:2593) — so ChatResponse.FinishReason is discarded and a finish_reason=length reply (token cap, provider cap, gateway truncation) is indistinguishable from a complete one.
2. The model just re-emits less. Asked to re-emit a long table verbatim, models frequently emit a subset; nothing checks that either.

Either way the shortened body overwrites the page. Because the overwrite is a normal page update it leaves no trace in the page history, and because a KB-wide re-ingest (triggered by a change to any document) rewrites unrelated pages too, a single ingest can quietly shrink pages nobody touched.

Measured on our KB: 156 pages are shorter than before a bulk ingest, 88,556 characters in total; one certificate-ledger page went from 23 holder rows to 4.

We have a fix + regression test ready (refuse a rewrite that drops table rows when nothing was retracted, keep the existing page and log it loudly). Happy to send it as a PR.

### Steps to Reproduce

1. KB with the wiki enabled, containing a page with a long markdown table (e.g. a ledger with dozens of rows) or a long prose body.
2. Run any ingest batch on that KB. It does not have to touch that page — the KB-wide re-ingest rewrites unrelated pages as well.
3. Compare wiki_pages.content before and after:
   select slug, length(content) from wiki_pages where knowledge_base_id = '<kb>' order by length(content) desc;
   Rows/characters drop, and the page's edit history shows nothing.

Deterministic reproduction (no model needed): stub the chat model to return a shortened rewrite (or FinishReason: "length" with a truncated body) and run the reduce path for a slug whose page already has content — the page is overwritten with the shorter body.

### Expected Behavior

A rewrite that loses content without a matching retraction is rejected: the existing page is kept, the batch logs it loudly ("refusing truncated rewrite ... keeping the existing page"), and the new information for that batch is reported as dropped instead of silently replacing good data.

### Actual Behavior

The rewrite is written back unconditionally (wiki_ingest_batch.go:2067), so a truncated or lazily-shortened reply destroys the previous body. No log line, no diff, no error: the only observable is that the page got shorter.

### WeKnora Version

App Version: v0.8.0 (Docker image revision 1edcd54b43606d9079bb36650efe3f68707a79ea) UI Version: v0.8.0 Verified still present on main: wiki_ingest_batch.go has no truncation guard around the page.Content = updatedBody write, and wiki_ingest.go:2593 still returns only response.Content, dropping finish_reason.

### Deployment Method

Docker Compose

### Operating System

macOS 26.6.2 (build 25G83, Apple Silicon), Docker Desktop

### Relevant Logs

```shell
There is nothing in the logs — that is the point. The write path has no failure to log:

$ docker logs WeKnora-app | grep -iE "truncat|refus"     # (nothing)

The only evidence is the data: page bodies are shorter than before the ingest, with no page-history entry.
```

### Confirmation

Bruce-Yii:
Fixes #3468.

`generateWithTemplate` returned `response.Content` while discarding
`response.FinishReason`, and the batch write-back stored any non-empty
rewrite unconditionally — so a token-budget-truncated rewrite
(`finish_reason=length`) silently shrank the page.

This change fails fast on `finish_reason=length` (warn + error, no
retry: the same prompt truncates deterministically), letting the
existing batch error path keep the original page. Follows the existing
`isTruncated` doctrine in `memory/extract.go`.

Validation: new `TestGenerateWithTemplateRejectsLengthTruncatedRewrite`
(red on unpatched code, green with guard) +
`TestGenerateWithTemplateAcceptsStopFinishReason`; `TestGenerateWithTemplate*`
5/5 green, `go vet`/`gofmt` clean on a gcc/cgo runner.

Base: Tencent/WeKnora@9aa63c866ff4570767110a6c7d3cf477a4ab955b. No behavior change for complete (`stop`) rewrites.


lyingbug (commented):
Thanks. Stopping a truncated page rewrite from overwriting the page is the right fix for #3468, and we'd like to merge it after two changes:

1. The guard lives in the shared `generateWithTemplate` (`internal/application/service/wiki_ingest.go`), so it affects every wiki LLM call, not just the page rewrite. The other callers are the document summary (`wiki_ingest_batch.go` ~1329), extraction, dedup, index intro, taxonomy planning, and the citation/candidate-slug calls. The summary is the clearest regression. Today a long document whose summary hits the output cap still gets a (slightly clipped) summary page. With this change, `summaryErr` is set, the whole document is sent to `failedOps` and requeued, and because the truncation is deterministic it fails the same way on every retry until the fail_count budget dead-letters it. Please scope the check to the page-rewrite path (`WikiPageModifyUserPrompt` at `wiki_ingest_batch.go` ~2042). Two ways to do that: add a variant or option that returns the finish reason, or return a typed/sentinel error (e.g. `errWikiLLMTruncated`) that only the rewrite caller treats as fatal while other callers keep the content.
2. Checking only `== "length"` misses equivalent values. `openaicompletions` passes `finish_reason` through verbatim, and some OpenAI-compatible gateways report `max_tokens` / `max_output_tokens`. Please match the same set as `isLengthFinishReason` in `chat_pipeline/chat_completion_stream.go` (case-insensitive, trimmed): `length`, `max_tokens`, `max_output_tokens`. A small helper in the service package, or an exported shared one, would do.

Bruce-Yii:
Both points addressed in abbd9170 (appended to this branch, no rebase):

1. Truncation guard is now rewrite-only: the shared generateWithTemplate just surfaces the finish reason and keeps returning content as before; a new generatePageRewrite wrapper fails fast with a typed sentinel (errWikiLLMTruncated), and only the WikiPageModifyUserPrompt call site (wiki_ingest_batch.go) uses it. Summary/extraction/dedup/intro/taxonomy/citation paths keep truncated output and stay out of failedOps.
2. Detection now matches isLengthFinishReason semantics (trim + case-insensitive): length, max_tokens, max_output_tokens, via a small service-local helper with table tests.

Validation: gofmt/go vet clean; focused tests pass (rewrite rejects all three reasons incl. case/padding variants, truncated summary returns content with nil error, stop unchanged); full service-suite failures identical to pristine (13 pre-existing env failures).

Bruce-Yii:
Hi @lyingbug,

I noticed #3492 was closed after the two review points were addressed in
abbd9170 (rewrite-only truncation handling + finish-reason normalization),
while #3468 is still open.

Could you let me know what prompted the close? If the preferred direction
or scope has changed, I’m happy to adjust the patch or re-cut it
accordingly.

Thanks.

On Tue, 22 Sep 2026 21:17:40 -0700, lyingbug ***@***.*** wrote:

Closed #3492 <https://github.com/Tencent/WeKnora/pull/3492>.

—
Reply to this email directly, view it on GitHub
<https://github.com/Tencent/WeKnora/pull/3492?email_source=notifications&email_token=CHDJZCZ2SIJQCDJE3SZOXAT5QNFGJA5CNFSNUABQM5UWIORPF5TWS5BNNB2WEL2JONZXKZKFOZSW45CON52GSZTJMNQXI2LPNYXTGMJWGQ4TONJRGI2TNJTSMVQXG33OUZQXK5DIN5ZKKZLWMVXHJLDGN5XXIZLSL5RWY2LDNM#event-31649751256>,
or unsubscribe

PR #3495, opened 2026-09-21
fix: remove temp file when TempFileContext write fails inside __enter__

Linxiushen:
## Description

`docreader/utils/tempfile.py` — `TempFileContext.__enter__()` creates a `NamedTemporaryFile(delete=False)` and then calls `write()` / `flush()` with no error handling:

```python
self.temp_file = tempfile.NamedTemporaryFile(suffix=self.suffix, delete=False)
self.temp_file.write(self.file_content)
self.temp_file.flush()
```

If either raises (ENOSPC, quota, any I/O error), the exception escapes `__enter__`. Python's `with` statement never calls `__exit__` for a context manager whose `__enter__` failed, and `__exit__` is the only place the file is closed and removed — so the file leaks permanently. `delete=False` means GC won't remove it either.

`docreader/parser/doc_parser.py:121` uses this to spill every uploaded `.doc`'s full bytes to disk before parsing, so one disk-pressure event leaves the whole document behind and makes the pressure worse.

**Fix:** wrap `write`/`flush` in `try/except`, close and remove the file, re-raise. The success path and the "exception inside the `with` body" path (which already goes through `__exit__`) are unchanged.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test

## Related Issue

None — found while reading `docreader/utils`. Happy to open one first if you prefer that flow.

## Testing

Failure injection (monkeypatch `write()` to raise `OSError(28, "No space left on device")`):

```
before: __enter__ raises OSError; created temp file still on disk after exception  -> LEAKED
after:  __enter__ raises OSError; created temp file removed                        -> cleaned up
```

Well-formed input, before vs after — identical: 400 KB payload with all 256 byte values, `exists=True`, `roundtrip_identical=True`, `suffix=".doc"`, `deleted_after_with=True`. Exception raised inside the `with` body still cleans up and propagates.

New `docreader/tests/test_tempfile_context.py` (3 tests). The failure-injection test **fails on `main`** and passes with this change.

Full docreader suite, `PYTHONPATH=. python -m pytest docreader/tests -q`:

```
main:        15 failed, 170 passed, 12 skipped
this branch: 15 failed, 173 passed, 12 skipped
```

The 15 failures are identical on both (diff of the failure sets is empty) and unrelated to this change — all in `test_excel_parser.py` (14) and `test_ssrf_proxy.py::TestSSRFProxy::test_webkit_redirects_and_subresources_use_proxy` (1), which look environment-dependent on my macOS box. I could not do an editable install of `docreader` (setuptools flat-layout error, "Multiple top-level packages discovered"), so I installed the declared dependencies into a venv and ran with `PYTHONPATH`.

## Checklist

- [x] Changed source files are formatted — no formatter is configured in `docreader/pyproject.toml`; I kept the file's existing style
- [x] Diff-scoped lint passes where applicable — Go-only item, N/A for this Python change
- [x] Self-reviewed the code
- [x] Updated related documentation — N/A, no user-facing behaviour change

---

*Investigated and fixed with AI assistance (Claude); I reviewed the change and ran the verification above myself.*


docreader/utils/tempfile.py
```diff
@@ -21,8 +21,17 @@ def __enter__(self):
...
-        self.temp_file.write(self.file_content)
-        self.temp_file.flush()
+        # If write/flush fails (e.g. ENOSPC), __exit__ is never called because
+        # the exception escapes __enter__, so clean up here or the file leaks
+        # permanently (delete=False means GC will not remove it either).
+        try:
+            self.temp_file.write(self.file_content)
+            self.temp_file.flush()
+        except BaseException:
+            self.temp_file.close()
```
lyingbug: `close()` can re-raise the flush ENOSPC (CPython buffered files flush again on close). With `delete=False`, that skips `os.remove()` and the temp file still leaks — the exact disk-pressure case this PR targets.

Please keep unlink in a path that always runs, e.g. swallow `OSError` from `close()`/`remove()` and then re-raise the original exception. A test should make `flush()` fail and `close()` raise so this is covered.

lyingbug (commented):
Review follow-up: `__enter__` cleanup is not safe on the ENOSPC path this PR is meant to fix.

`NamedTemporaryFile.close()` flushes again. After `write()`/`flush()` fails with ENOSPC, `close()` typically re-raises, so `os.remove()` never runs and the file still leaks. The existing test mocks `write()` to raise immediately, so it does not hit this path.

Pushed a follow-up commit that shares `_cleanup()` (always unlink, swallow cleanup `OSError`, re-raise the original error) and adds a test where `flush()` fails and `close()` raises a different error.

Linxiushen:
Thanks — you're right, and the follow-up is the correct shape. I reviewed 50c69da and checked it locally:

- `docreader/tests/test_tempfile_context.py`: 4 passed on the branch head; the new `test_removes_file_when_close_reraises_after_flush_fails` fails against my original ef54ced (the `close()` re-raise skipped `os.remove`), so it covers exactly the gap.
- Independently confirmed the premise with a plain `io.BufferedWriter` over a raw stream that starts returning ENOSPC: after `flush()` raises errno 28, `close()` flushes again and re-raises errno 28 — so the previous `close(); os.remove()` ordering did leak on the path this PR targets.

Nothing further from my side; the branch is ready as it stands.

PR #3510, opened 2026-09-21
fix(container): provide access.SharedAgentLookup so the app can boot

lvchenyang:
## Description

`a3c7be10` (#3475) added an `access.SharedAgentLookup` parameter to **both** `NewSkillHandler` and `NewMCPServiceHandler`, but only wired the skill handler in the container:

```go
must(container.Provide(handler.NewMCPServiceHandler))              // line 540, unchanged
must(container.Provide(func(
    s *service.TenantSkillService, agents interfaces.AgentShareService,
) *handler.SkillHandler {
    return handler.NewSkillHandler(s, s, agents)                   // line 552, adapted
}))
```

`dig` matches by exact type. Nothing provides `access.SharedAgentLookup` — only `interfaces.AgentShareService` — so `*handler.MCPServiceHandler` cannot be built, `NewRouter` fails, and the binary dies on every start:

```
FATAL main.go:141[main] | Failed to run application: could not build arguments for
function "main".main.func1 (/app/cmd/server/main.go:65): failed to build *gin.Engine:
could not build arguments for function ".../internal/router".NewRouter (router.go:98):
failed to build *handler.MCPServiceHandler: missing dependencies for function
".../internal/handler".NewMCPServiceHandler (mcp_service.go:35):
missing type: access.SharedAgentLookup (did you mean interfaces.AgentShareService?)
```

This is reproducible on current `main`: build the app image from `4a9aa062` and it crash-loops before serving a single request.

The fix registers the narrow lookup once, next to the share service it is derived from, instead of wrapping each consumer:

```go
must(container.Provide(service.NewAgentShareService))
must(container.Provide(func(s interfaces.AgentShareService) access.SharedAgentLookup { return s }))
```

A handler that only resolves a shared agent should depend on the lookup, not on the whole share service, and the next consumer of `SharedAgentLookup` will not have to repeat the adapter. `NewSkillHandler`'s existing closure is left alone — it is there to pass `*service.TenantSkillService` twice for two different interfaces, which a plain `Provide` still cannot express.

`access.SharedAgentLookup` and `access.AgentShareLookup` are the only `access.*` types taken by a constructor anywhere in `internal/handler`, and the latter is not used in DI, so this is the last gap of its kind.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [ ] 🧪 Test

## Related Issue

No issue filed. Regression from #3475 (`a3c7be10`), still present on `main` at `4a9aa062`.

## Testing

Scoped to the changed package:

```
go build ./internal/container/          # ok
go vet ./internal/container/            # ok
go test ./internal/container/           # ok  1.911s
gofmt -l internal/container/container.go  # clean
git diff --check origin/main...HEAD     # clean
```

The closure body `return s` is itself the compile-time proof that `interfaces.AgentShareService` satisfies `access.SharedAgentLookup`.

**Not run locally:** `golangci-lint run --new-from-rev=origin/main ./...` — `golangci-lint` is not installed in this environment, leaving it to CI. The added lines are gofmt-clean and the longest is 103 columns at `tab-width: 4`, under the configured `lll` limit of 120.

**Note on coverage:** nothing under `internal/container/` exercises `BuildContainer`, so the dig graph is only resolved at run time. That is why this passed compilation, the package tests and CI, and only surfaced on boot.

*Corrected after opening this PR:* I originally wrote here that covering `BuildContainer` would mean standing up the full dependency graph (database, Redis, gRPC, object storage), and used that to justify shipping the fix without a test. That is not true. `dig.DryRun(true)` resolves the entire graph without constructing anything, which makes such a test cheap. #3511, opened independently a few minutes after this one, does exactly that; I have verified its `TestRouterWiring` fails on `main` at `4a9aa062` with this PR's exact error and passes against this PR's wiring, in ~2s with no external services. See the comment below on how the two fit together.

## Checklist

- [ ] Diff-scoped lint passes where applicable — not run locally, see Testing
- [x] Self-reviewed the code
- [ ] Added/updated tests covering the change — see "Note on coverage" above
- [x] Updated related documentation (README, `docs/`, Swagger annotations, etc.) — N/A


lvchenyang:
Duplicate of #3511 — same crash, opened independently about three minutes after this one.

The two PRs split cleanly, and each has the better half of the other:

**Wiring.** This PR registers `access.SharedAgentLookup` once, next to the `AgentShareService` it is derived from (+4). #3511 wraps `NewMCPServiceHandler` in a closure forwarding all five parameters (+11). Both unbreak the boot. The difference only shows up for the *next* consumer: `NewSkillHandler` already needed this type and got its own adapter in a3c7be10, so a shared provider means a third consumer needs no wiring at all, while the per-handler closure has to be repeated.

**Test.** #3511 adds `router_wiring_test.go`, which resolves the production graph under `dig.DryRun(true)` and asserts `*gin.Engine` builds. This PR has no test, and the reason I gave for that was simply wrong — I have corrected the description.

I checked the combination locally. #3511's `TestRouterWiring`, applied on its own to `main` at `4a9aa062`, fails in both `lite` and `redis` modes with exactly:

```
resolve router dependencies: ... failed to build *handler.MCPServiceHandler:
missing type: access.SharedAgentLookup (did you mean interfaces.AgentShareService?)
```

and passes against this PR's wiring, in ~2s with no database, Redis or gRPC. So the two compose without conflict.

What seems best to me is the provider from this PR plus @lyingbug's test — but **either PR alone unbreaks `main`**, which is currently crash-looping on boot for anyone who builds the app image from `4a9aa062`, so please just take whichever is quicker to land. I am equally happy to close this in favour of #3511, or to add the graph test here with credit. The test is @lyingbug's idea either way, and it is the part that stops this recurring.

lvchenyang:
#3511 has landed as 2d37f8fd, so `main` boots again. Closing this in favour of it — the graph test it brings is the part that stops this recurring, and that is @lyingbug's.

If the shared provider from this PR is ever wanted (so that a third consumer of `access.SharedAgentLookup` needs no wiring of its own), it is a 4-line follow-up on top of the merged fix, and `TestRouterWiring` already covers it. Happy to open one if a maintainer asks.

PR #3518, opened 2026-09-21
fix(wiki): include exact alias matches in dedup candidates

autumnseasonism:
## Description

Follow-up to #3418: include exact matches against existing Wiki page aliases when retrieving deduplication candidates, rather than relying on title similarity alone.

For example, an existing page titled `国际商业机器公司` with alias `IBM` should be offered as a candidate for a newly extracted `IBM`, even if its title does not pass the trigram threshold. This changes candidate retrieval only; an alias match does not automatically merge entities.

## Changes

- Match aliases case-insensitively, trimming surrounding spaces; keep punctuation significant and avoid alias substring matching.
- Preserve title similarity matching and the knowledge-base, page-type, archived-page and soft-delete filters.
- Score exact alias matches at 1 and use slug ordering for deterministic ties; preserve the default and maximum limits.
- Query title and alias candidates through separate bounded branches, then deduplicate by slug and apply the final top-k limit. This preserves the title index path and avoids the combined OR query's overhead in measured selective-query cases.
- Add an opt-in native PostgreSQL regression test for alias-only matches, normalization, scope filters, shared/repeated aliases, title/alias overlap, bounded top-k ordering, null arrays, empty/SQL-looking input and limits.

No schema migration, new production dependency, prompt change or historical data cleanup is included.

## Type of Change

- [x] Bug fix
- [x] Test

## Related Issue

Follow-up to #3418; no issue is automatically closed by this PR.

## Testing

Tested the actual project source at f574a43 on Linux with Go 1.26.1, CGO enabled, GCC/G++ 13 and PostgreSQL 16.15 with pg_trgm (not just an extracted-function harness).

- PASS: new PostgreSQL regression test, three consecutive runs, each with 13 named cases plus default/max-limit checks.
- PASS: all repository-package tests — 225 top-level tests / 281 including subtests; no failures or skips.
- PASS: Wiki-related service tests — 134 top-level tests / 234 including subtests; no failures or skips.
- PASS: `go vet ./internal/application/repository ./internal/application/service`.
- PASS: `go build ./cmd/server` (linker emitted PIE/DT_TEXTREL warnings).
- PASS: formatting and diff checks.

The full service-package test run was **not all green** in the local environment: both this branch and the unchanged base cc82e75 had 2331 passes, the exact same 40 failures, and 1 existing skip (counts include subtests). Observed failures involve outbound URL safety checks rejecting domains such as example.com and api.e2b.app that the local network resolves to 198.18.x.x. No safety checks were disabled. The existing skip is TestCreateStore_DifferentEndpointSameIndex_Allowed. A clean CI run with normal DNS is still needed. The entire repository test suite and end-to-end LLM behavior were not run.

To run the added regression against a dedicated test PostgreSQL database with pg_trgm installed:

```sh
WEKNORA_TEST_POSTGRES_DSN='host=127.0.0.1 port=5432 user=postgres dbname=postgres sslmode=disable' \
  go test -count=3 -v ./internal/application/repository -run '^TestFindSimilarPages_ExactAliasesPostgres$'
```

Without the DSN, this PostgreSQL-specific test skips. Fixtures use a transaction-local temporary table and roll back; they do not modify persistent Wiki data.

## Performance / limitations

Exact alias matching still scans JSONB arrays within the scoped knowledge base. This adds work compared with upstream title-only lookup and is not an indexed-alias solution.

In a separate native PostgreSQL 17.11 synthetic benchmark (nine rounds with variant order rotated), splitting the branches reduced median execution time versus an initial combined-OR implementation from 30.2 to 13.3 ms for a 10k-page alias hit and from 124.4 to 51.7 ms for a 40k-page alias hit, with identical ordered candidates. Broad title queries were approximately unchanged; small datasets could be slightly slower. These are local EXPLAIN ANALYZE measurements with a reduced slug/score projection, not production latency guarantees or a claim of improvement over title-only retrieval. Representative production performance remains to be validated.

## Checklist

- [x] Targeted repository and Wiki service tests pass; unrelated full service-suite failures documented above
- [x] Diff-scoped golangci-lint (not run locally; go vet passed)
- [x] Unrelated/environment-dependent broader-check failures and unrun checks are documented above
- [x] Self-reviewed the code
- [x] Updated interface documentation
- [x] No breaking API or schema change


lyingbug (commented):
Thanks — surfacing exact alias hits as dedup candidates is the right fix, and the write-up of the measurements is very helpful. One thing before we merge:

1. **Per-term cost multiplies across an ingest batch.** The caller (`internal/application/service/wiki_ingest.go`, `probe` around L2340–2360) calls `FindSimilarPages` sequentially for the item name *and every alias*, for every extracted entity/concept. With the alias branch now scanning all entity/concept pages' JSONB `aliases` in the KB, your ~50 ms at 40k pages becomes one scan per term — a batch with a few dozen items × several aliases can add seconds per chunk. Please either:
   - batch the alias side: one query per ingest batch using `lower(btrim(alias.value)) = ANY(?)` that also returns the matched term, then map hits back to items in Go (keeps title lookups as they are); or
   - add an index that can serve the lookup (e.g. a normalized aliases column/generated array with a GIN index), or
   - at minimum document the expected KB size bound in the `FindSimilarPages` comment.
   The first option seems the least invasive.
Optional / non-blocking:
- `TestFindSimilarPages_ExactAliasesPostgres` skips unless `WEKNORA_TEST_POSTGRES_DSN` is set, and no CI workflow sets it, so this path currently has no CI coverage. If you see a cheap way to exercise the query by default, that would be welcome; otherwise please mention in the test comment that it is manual-only.

autumnseasonism:
Thanks for pointing out the per-ingest cost. Both suggestions are addressed in `83cfba6` and `3aa532a`:

1. **Batch candidate retrieval.** Ingest now deduplicates normalized names/aliases across entities and concepts and calls `FindSimilarPagesBatch` once. A materialized CTE expands scoped aliases once and returns the matched term. Title probes keep their trigram predicate and independent top-k inside per-term lateral branches; aliases and titles are combined with the same score/slug ordering. Go maps each term's hits only to the items that supplied it, preserving the cross-item merge guard. The single-term method remains a wrapper, and an alias hit still does not automatically merge anything.
2. **Real PostgreSQL coverage in CI.** App now provisions a disposable PostgreSQL 16 service with pg_trgm and explicitly runs `TestFindSimilarPages_ExactAliasesPostgres`. A required-DSN flag plus a named-test JSON result check rejects skipped, failed or zero-selected tests. Local runs without a DSN remain optional. No production database or repository secrets are used.

Validation on the actual project: [App run 35733563572](https://github.com/Tencent/WeKnora/actions/runs/35733563572) **passed**, including vet/tests, server build, and the new PostgreSQL step. All 14 named database subtests passed, including batch ownership/query-count and per-term limits; the anti-skip check returned true. New service tests cover shared terms, item ownership, duplicate/empty candidates and lookup failure. Local actionlint and positive/negative anti-skip checks also passed.

I also compared full-projection, ordered candidates with the pre-batch implementation on synthetic PostgreSQL 17.11 data (three alternated rounds, analyzed tables, GIN pending list cleaned after bulk insertion). At 40k pages / 40 unique alias terms, median whole-batch wall time changed from **2957.61 ms to 79.06 ms**, with identical candidates; 120 terms changed from 9822.09 ms to 97.25 ms. EXPLAIN showed one alias CTE build, JSON expansion once per scoped page, and indexed title probes. These are local synthetic results, not production guarantees. Single-term queries may be slightly slower; dense shared aliases, very large batches and title-index pending-list costs still matter. No alias index/schema migration is added. Details and limitations are in the updated PR description.

One unrelated check is still red: [Go Lint run 35733563641](https://github.com/Tencent/WeKnora/actions/runs/35733563641) reports eight line-length findings in `internal/handler/knowledge_spans_test.go` (193–195, 229–230, 258–259, 266), outside this PR's changed files. The run checks synthetic merge `ce7f864` relative to old base `cc82e75`; that handler file is byte-identical to upstream parent `0a99a28` (blob `782baead704c8feb0e8e4dbfd67c113f8ec30ca6`). I have left unrelated upstream code and lint settings untouched. Could you advise how you prefer to handle that upstream/check-baseline issue?

autumnseasonism:
Follow-up on the unrelated lint failures: `3e892c1` pins the golangci job checkout to `github.event.pull_request.head.sha`. The workflow already documents that it judges only PR-introduced code, but the default synthetic merge checkout included newer main changes while the diff filter used the older PR base. That combination attributed the eight upstream handler findings to this PR.

All lint rules, the merge-base filter and module detection remain enabled. App CI still tests the merged tree for integration compatibility. No handler code was changed or findings suppressed. The workflow change also triggers lint for all three modules, not fewer checks.

All six checks on head `3e892c1` are now green:

- [Go Lint](https://github.com/Tencent/WeKnora/actions/runs/35734636049): root, cli and client all pass, each reporting 0 issues; module detection also passes.
- [App](https://github.com/Tencent/WeKnora/actions/runs/35734635965): vet/tests, build and required PostgreSQL regression all pass on the merged tree.
- Repository scan passes.

This supersedes the red-check status in my previous reply. The PR description is updated with the final run links. Ready for your next review; no merge has been performed.

PR #3522, opened 2026-09-21
fix(agent): emit the knowledge references an agent turn retrieved

### LINKED ISSUE #3501: agent-chat 缺少引用回传：无 references 事件、answer 无 <kb> 标签，citation_enabled 开启无效
### Affected Component

Frontend UI

### Bug Description

外部平台通过 API 集成 WeKnora 时，**/api/v1/knowledge-chat** 端点能正常回传引用：SSE 流中有 `references` 事件（带 `knowledge_references` 列表），answer 文本内嵌 `<kb>` 标签，下游可渲染引用角标与溯源链接。

但同一个自定义智能体改走 **/api/v1/agent-chat** 端点时，回答没有引用信息：SSE 流中始终没有 `references` 事件，answer 文本中也没有 `<kb>` 标签。在智能体配置中设置 `citation_enabled: true` 后行为无变化。

这使得 agent 模式（ReAct + MCP 工具 + 知识库）下，基于 knowledge_search 检索结果生成的回答无法被下游平台渲染引用角标，回答内容无法溯源。

### Steps to Reproduce

1. 创建自定义智能体（POST /api/v1/agents）：config 设 kb_selection_mode=selected 并绑定知识库（knowledge_bases）、allowed_tools 含 knowledge_search、rerank_model_id 已配置、citation_enabled=true
2. POST /api/v1/agent-chat/{session_id}，body: {"query":"<知识库可回答的问题>","knowledge_base_ids":["<kb_id>"],"agent_id":"<agent_id>","channel":"api","stream":true}
3. 观察 SSE 事件流类型与 answer 文本
4. 对照：同一问题走 POST /api/v1/knowledge-chat/{session_id}，可观察到 references 事件与 answer 内嵌 <kb> 标签

### Expected Behavior

agent-chat 在回答引用了知识库检索内容时，同样回传 references 事件（knowledge_references 列表）或在 answer 中内嵌 <kb> 标签，格式与 knowledge-chat 保持一致，便于外部集成平台渲染引用角标并做溯源跳转。

### Actual Behavior

agent-chat 的 SSE 事件序列实测为 agent_query ×1 / session_title ×1 / tool_call ×N / tool_result ×N / answer ×N / complete ×1，始终没有 references 事件；answer 为纯文本，无 <kb> 标签。citation_enabled=true 前后行为一致。多次重试（含工具完整调用的样本）结果相同。

### WeKnora Version

v0.8.0（docker-compose 部署）

### Deployment Method

Docker Compose

### Operating System

Linux x86_64（CentOS 7.9，Docker 24.x）

### Relevant Logs

```shell
SSE 事件序列对比（同一智能体、同一问题）：

agent-chat: {"agent_query":1,"session_title":1,"tool_call":16,"tool_result":6,"answer":159,"complete":1}  ← 无 references
knowledge-chat: {"agent_query":1,"tool_call":1,"tool_result":1,"references":1,"thinking":N,"answer":N,"complete":1}  ← 有 references + answer 内嵌 <kb doc=".." chunk_id=".." kb_id=".."/>
```

### Confirmation

vinhchuquang:
## Description

`agent-chat` answers from the knowledge base and cites nothing. The reported
symptom is a missing `references` event, but the cause is one field that is
never written.

`types.AgentState.KnowledgeRefs` is documented as *"Collected knowledge
references"*. It is declared in `internal/types/agent.go`, initialised empty in
`engine.go`, read by `finalize.go` into the `complete` event — and there is no
`append` to it anywhere in the tree:

```
$ grep -rn KnowledgeRefs internal/agent internal/types
internal/agent/engine.go:350:   KnowledgeRefs: []*types.SearchResult{},
internal/agent/finalize.go:156: knowledgeRefsInterface := make([]interface{}, 0, len(state.KnowledgeRefs))
internal/agent/finalize.go:157: for _, ref := range state.KnowledgeRefs {
internal/agent/finalize.go:167:         KnowledgeRefs:   knowledgeRefsInterface,
internal/types/agent.go:484:    KnowledgeRefs        []*SearchResult `json:"knowledge_refs"`
```

The ReAct loop runs `search_knowledge`, gets its chunks, renders them for the
model, and drops them.

### The blast radius is wider than agent-chat

`EventAgentReferences` has exactly one emitter — `emitKnowledgeReferencesEvent`
in `session_knowledge_qa.go`, the knowledge-chat pipeline. It has four
subscribers:

| Subscriber | Consequence today |
|---|---|
| `handler/session/agent_stream_handler.go:144` | no `references` SSE event — the reported symptom |
| `im/service.go:2614` | IM agent replies carry no sources |
| `im/service.go:3031` | same on the second IM delivery path |
| `mcpserver/tools_ask.go:256` | the MCP `ask` tool returns no references |

All four also read `KnowledgeRefs` off the `complete` event, which is empty for
the same reason. So every agent-mode surface loses citations, not just the SSE
endpoint in the issue.

### Change

`ToolResult` gains a live-only `KnowledgeRefs []*SearchResult` with `json:"-"`,
next to the existing `OutputFiles` precedent: the chunks stay out of SSE replay
and `agent_steps` storage, where they would duplicate what the assistant
message already persists. `formatOutput` fills it — the `Data` rows it already
builds are flattened for the model and drop fields the reference panel needs,
so they cannot be turned back into a `SearchResult` without guessing.

A new `collectKnowledgeRefs` runs once per round, right after the round's tools
have finished, and does two things: appends the new hits to
`state.KnowledgeRefs`, and emits `EventAgentReferences` with them.

Two properties the existing consumers require, both covered by tests:

- **Each chunk is announced once.** `collectIMKnowledgeReferences`,
  `collectReferences` and `handleReferences` all *append* what they receive
  with no de-duplication, so a chunk that two rounds both retrieved would
  otherwise show up twice.
- **One event per round, not per tool call.** Parallel calls finish in
  arbitrary order but are appended to the step in model order, so batching at
  the round boundary keeps citation order stable.

The `complete` handlers all *replace* rather than append
(`msg.KnowledgeReferences = types.References(refs)`), so the round events and
the final event do not double up.

### Scope

- The collector matches on `ToolResult.KnowledgeRefs` being non-empty, not on
  the tool name, so a second retrieval tool needs no change here.
- A turn is capped at 200 distinct references. `search_knowledge` returns up to
  30 chunks per call and `MaxIterations` can be 30 with several searches per
  round, so an uncapped list would let one assistant message persist megabytes
  of chunk text — a regression this change would have introduced, since the
  list was always empty before.
- **Inline `<kb>` tags are not addressed.** They come from a different
  mechanism that is already wired in the agent path
  (`modelcontext.NewRegistry` at `engine.go:124`, `RegisterContextChunk` at
  `observe.go:829`), so a missing tag there has its own cause. The issue's
  Expected Behavior asks for the `references` event *or* the tags; this
  delivers the event and the `knowledge_references` list.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🎨 Refactor
- [x] 🧪 Test

## Related Issue

Fixes #3501

## Testing

Six new tests. `internal/agent/references_test.go` pins the collector against
what the four subscribers actually do:

```
TestCollectKnowledgeRefsAnnouncesEachChunkOnce   three rounds re-retrieving the same
                                                 chunks announce {a,b} then {c} then nothing
TestCollectKnowledgeRefsBatchesARound            two parallel calls -> one ordered batch
TestCollectKnowledgeRefsStaysSilentWithoutHits   nil results, non-retrieval tools and nil
                                                 refs emit nothing rather than an empty panel
TestCollectKnowledgeRefsStopsAtTheTurnCap        225 hits over three rounds -> 200 kept,
                                                 the round with no room stays silent
```

`internal/agent/tools/search_knowledge_references_test.go` pins that
`formatOutput` carries the typed results, and that a no-hit search carries none.

`go build ./...`, `go vet ./internal/agent/... ./internal/types/...
./internal/im/... ./internal/mcpserver/... ./internal/handler/...`, `gofmt -l`
and `gofumpt -l` are all clean.

`go test ./internal/agent/... ./internal/types/... ./internal/mcpserver/...
./internal/modelcontext/...`: all pass except `internal/agent/tools`, which
fails five sandbox and shell cases that need a POSIX environment
(`TestShellStdinPreservesLiteralDataForTheEntireCommand`,
`TestSkillPythonPackageRecoveryInstallsWithoutPip`,
`TestReadFileCombinesSourcesWithoutGrantingHostAccess`,
`TestMCPSchemaDoesNotLoadExternalResources`,
`TestSkillPackageCommandsRefuseMissingDirectory`).

`go test ./...`: 108 packages `ok`, 9 packages failing — `agent/tools`,
`application/service`, `database`, `handler`, `localsandbox`,
`localsandbox/core`, `localsandbox/seatbelt` (macOS-only), `models/catalog`
and `sandbox`. That is a Windows baseline, not a regression: running the same
nine packages from a clean `origin/main` checkout of this worktree fails the
identical set of test names.

```
branch: 82 failing tests
main:   82 failing tests
=== DIFF (empty = identical) ===
IDENTICAL
```

No package this change touches appears in that set beyond `agent/tools`, whose
five failures are the POSIX ones listed above and are present on `main` too.

Diff-scoped lint, the same filter CI uses:

```
$ golangci-lint run --new-from-merge-base=origin/main ./...
0 issues.
```

## Checklist

- [x] Diff-scoped lint passes where applicable (`--new-from-merge-base=origin/main` — `0 issues`)
- [x] Self-reviewed the code
- [ ] Updated related documentation — no user-facing doc describes the references event; the behaviour now matches what `knowledge-chat` already documents

## Screenshots / Recordings

Not applicable — no UI change. The reference panel is existing frontend that
renders the `references` event it was already listening for.


lyingbug (commented):
Thanks for tracking this down. Wiring `KnowledgeRefs` back up is the right fix, and we'd like to merge it after one small change:

1. `internal/agent/references.go`, in `collectKnowledgeRefs`: the event ID is currently `fmt.Sprintf("round-%d-references", state.CurrentRound)`, so it repeats every turn. Round 1 of every agent turn in a session emits `round-1-references`. Every other agent event gets a unique ID from `generateEventID(...)` (`internal/agent/const.go`), and so does the existing KnowledgeQA references emitter (`generateEventID("references")` in `session_knowledge_qa.go`). The stream handler forwards `evt.ID` as-is, and the frontend uses `data.id` to find the assistant row in `resolveActiveAssistantMessage` (`useChatStreamHandler.ts`). Because the ID repeats, a references event can end up bound to an earlier turn's row. Please use `generateEventID("references")` here to match the other emitters.

vinhchuquang:
Thanks — you're right, the id repeated across turns. Switched to `generateEventID("references")` in e631331, and added `TestCollectKnowledgeRefsGivesEachEventAUniqueID`, which runs two turns that both start at round 1 and fails on the old id with `Should not be: "round-1-references"`.

PR #3529, opened 2026-09-22
feat(tracing): support custom metadata via X-Langfuse-Metadata header

liyihang866:
## Description

Allow callers of WeKnora's traced endpoints to attach arbitrary business
key/value pairs to the Langfuse trace metadata via a JSON request header.

When an upstream system (e.g. an IT service desk) calls a traced endpoint, it
can now tag the generated trace with its own business labels for filtering and
cost attribution in the Langfuse UI:

```http
POST /api/v1/knowledge-chat/:session_id
X-Langfuse-Metadata: {"ticket_id":"T-123","biz_line":"order","priority":"P1"}
```

The header value is parsed as a JSON object and merged into the request's
Langfuse trace metadata (`langfuse.trace.metadata`). Design decisions:

- **Additive, non-clobbering** — built-in correlation fields set by
  `GinMiddleware` (`http.method`, `http.path`, `http.query`, `request_id`,
  plus finish-time `status` / `response.size`) always win on conflict, so
  callers cannot overwrite internal diagnostics.
- **Fail-safe** — malformed JSON, non-object JSON, or an empty/absent header
  are silently ignored and never fail or delay the request.
- **Zero-cost when unused** — omitting the header keeps behaviour identical
  to the current release; no allocation or parse happens.

This mirrors Langfuse SDKs' native `metadata=` parameter support
(`create_trace(metadata={...})`), exposed for WeKnora's HTTP surface without
requiring a custom SDK.

## Type of Change

- [x] ✨ New feature

## Related Issue

None (feature request; happy to open an issue if maintainers prefer tracking).

## Testing

Targeted tests added in `internal/tracing/langfuse/middleware_test.go`
(9 test cases, all pass):

- `TestMergeMetadataHeader` (7 subtests): empty/whitespace header, valid JSON
  merge, malformed JSON ignored, non-object JSON ignored, built-in fields win
  on conflict, and numeric/boolean values pass through.
- `TestGinMiddlewareMergesMetadataHeader`: a real Gin route with
  `X-Langfuse-Metadata` exports the merged keys on `langfuse.trace.metadata`
  while preserving built-in fields (verified against an in-memory OTel span
  exporter).
- `TestGinMiddlewareIgnoresMalformedMetadataHeader`: malformed header neither
  blocks the request nor injects keys, and the request is still traced.

Validation commands (matching the PR checklist):

```bash
git diff --check origin/main...HEAD   # clean
gofmt -l internal/tracing/langfuse/   # empty
go vet ./internal/tracing/langfuse/   # pass
go test ./internal/tracing/langfuse/ -count=1   # ok (all tests incl. existing)
go build ./internal/tracing/langfuse/  # pass (CGO on)
```

Additionally verified end-to-end against a live deployment: a
`knowledge-chat` request carrying `X-Langfuse-Metadata:
{"ticket_id":"T-META-999","biz_line":"order","priority":"P1"}` produced a
trace whose `traces.metadata` in ClickHouse contained all three custom keys
alongside the untouched built-in fields.

## Checklist

- [x] Self-reviewed the code
- [x] Updated related documentation (`docs/Langfuse集成.md` — new section 2.4)
- [x] Breaking changes are clearly called out in the description above (none — purely additive)

## Screenshots / Recordings

N/A (backend-only change, no UI).


lyingbug (commented):
Thanks — the header-based metadata merge is small, fail-safe, and useful for callers that want to tag traces, so we'd like to take it.

1. **Remove `docs/plans/2026-09-22-langfuse-business-metadata-enrichment.md`.** It describes a design that this PR doesn't implement (`LANGFUSE_METADATA`, `StaticMetadata`, "方案 A" backfill in `qa.go`/`asynq.go`), so it would mislead readers about what exists. If you plan to do that work, please open an issue for it instead.
2. **Move the documentation to `website-docs/03-features/16-observability.md`** and revert the change to `docs/Langfuse集成.md`. `docs/` is deprecated. A short subsection there with the header name, an example, the "built-in fields win" rule and the limits from point 3 is enough.
3. **Bound the header in `mergeMetadataHeader`** (`internal/tracing/langfuse/middleware.go`). Right now any caller can send a header of arbitrary size/key count and it's copied into `langfuse.trace.metadata` on every traced request. Please ignore the header when the raw value exceeds a fixed size (e.g. 4 KB) and cap the number of merged keys (e.g. 32), with tests for both.

Optional / non-blocking:
- Consider accepting only scalar values (string/number/bool) so nested objects don't end up in trace metadata.

PR #3533, opened 2026-09-22
refactor(search): unify the ranking core shared by chat and knowledge-search

### LINKED ISSUE #3093: Ranking core (compositeScore / applyMMR) is implemented twice — chat pipeline vs agent knowledge_search — and the two copies have diverged

vicfei:
## Description

Closes #3093 — second attempt after #3150 was closed as stale against the pre-#3381 layout (thanks for the pointer; the duplication it targeted is unchanged in the reworked files).

`compositeScore` and `applyMMR` are still implemented twice — chat pipeline (`rerank.go`) and agent knowledge-search (`search_knowledge.go`) — and the copies have drifted: the agent side multiplies in a position prior chat does not apply, carries a faster incremental MMR form, and holds three one-line delegates over `searchutil`.

This extracts the one implementation both sides already agree on into `searchutil`, keeping only what genuinely differs at each call site:

- **`CompositeScoreRaw`** — the unclamped `0.6*model + 0.3*base + 0.1*source` core (web_search → 0.95). The clamp stays with the callers because the agent multiplies in the prior first; that one behavioral difference is now a documented, centrally-located `PositionPrior` instead of a buried private copy. Whether chat should adopt the prior stays a separate decision, as discussed on #3093.
- **`ApplyMMR`** — the incremental maxRedundancy form over caller-injected token sets, returning selection plus the average pairwise redundancy chat logs. Chat keeps its `ParallelMap` tokenization and pipeline observability; the agent tool keeps its sequential enriched-passage tokenization.

## How I verified

- **Bit-identity with both originals**: `TestCompositeScoreRawParityWithBothPipelines` replays the two legacy formulas (kept verbatim as reference oracles) over a full matrix of scores × sources (incl. case variants) × positions — the unified paths reproduce both, exactly.
- **MMR equivalence**: a 300-round randomized differential compares the shared `ApplyMMR` against the naive per-round rescan for selection order **and** average redundancy; the agent package's existing naive-oracle test now exercises the wrapper over the shared core the same way and passes unchanged.
- `go build ./...`, vet clean; `searchutil`, `agent/tools`, `chat_pipeline`, and the full `application/service` suite green (only the documented host-Python environmental failures remain, identical on pristine main).

## Notes

- No weight, lambda, or logging behavior changes anywhere — λ stays the callers' 0.7, constants live in one place now.
- The unified core is shaped so a configurable weighted variant could layer on top later without re-fragmenting (the score inputs are already parameters).


lyingbug (commented):
Thanks for picking this back up — the extraction is clean and the parity tests make it easy to trust, so we'd like to merge it. A few small things first:

1. **golangci-lint is failing:** `internal/searchutil/ranking_test.go:15` — `ineffectual assignment to sourceWeight (ineffassign)` in `legacyChatComposite`. Declare it as `var sourceWeight float64` (or drop the `default` branch) so the reference copy still reads like the original but passes lint.
2. **Drop the hand-written `lowerASCII`** (`internal/searchutil/ranking.go`, ~L26–44). Avoiding the `strings` import isn't worth a custom case-folding helper; `strings.EqualFold(knowledgeSource, "web_search")` is equivalent here and needs no allocation. The test helper can use `strings.ToLower` like the original code did.
3. **Fix the `PositionPrior` doc comment** (`ranking.go`, ~L46–50). It says later chunks go "down to -0.05", but with the chunk's own `startAt/endAt` the ratio `1 - start/(end+1)` is always positive, so the result is never below 1.0.

Optional / non-blocking:
- Related to 3: the ratio works out to `chunkLen/(endAt+1)`, so every chunk starting within roughly the first 19 chunk-lengths gets the full +0.05, and beyond that the bonus depends as much on chunk size as on position. Keeping the existing behavior in this refactor is right; a follow-up issue to compute it against document length (or drop it) would be appreciated.

vicfei:
Thanks for the review — all three items are addressed at eea10f3d (rebased onto current main, CI green):

1. The ineffassign was fixed in 78977b5d later that day (we chased the same CI failure); the reference oracle now declares `var sourceWeight float64`.
2. `lowerASCII` is gone — `isWebSearchSource` uses `strings.EqualFold`, and the two legacy oracles in the test use `strings.ToLower` like the original code did.
3. The `PositionPrior` doc now states the actual range: with the chunk's own `startAt/endAt` the ratio is always positive, so the multiplier stays within [1.0, 1.05] — an early-in-span chunk gets the full +0.05 and there is no below-1.0 penalty.

The optional follow-up you suggested is filed as #3663: the ratio works out to chunkLen/(startAt+chunkLen), so any chunk starting within ~19 own-lengths of the document start gets the full +0.05 — in practice nearly every chunk — and where it does vary it tracks chunk size as much as position. Options sketched there for computing it against document length or dropping the bonus.

